PR #5, opened 2016-05-19
Align package name with pypi name

emonty:
If you were to pip install ansible-container, you wouldn't expect a
top-level package called container. (It's totally reasonable that there
might be another pypi project called "container") Rather than just
making this ansible.container, which would require namespace packages in
ansible itself, call it ansible_container so that people don't get a
surprise in their global namespace.


j00bar:
`docker-compose` installs a top level package `compose`.
Many Django PyPI package names that begin with `django-` leave off the Django at the start, e.g. `pip install django-social-auth` has a top level package `social_auth`.
Ansible Galaxy's top level package is `galaxy`.

¯\_(ツ)_/¯

emonty:
I would consider those all to be making choices that trample on namespacing personally. (also, galaxy isn't open source, so I'd never pip install it anyway.) Do people just not install django things globally on their computers?

But shrug. I don't feel strongly - it just seems weird to me to have a thing take over such a generic word as "container" in the top level of the global namespace.

PR #14, opened 2016-05-25
Update README.md

tbarr:
Changed how we talk about Dockerfile and removed the word "droppings".


j00bar:
But I _love_ calling them "Ansible droppings" 😡 But you're right, you're right.... 🙄

PR #19, opened 2016-06-01
Modularize container engine/orchestrator

j00bar:


j00bar:
Merging this now. Gonna keep doing feature work.

PR #36, opened 2016-06-07
Fix shipit command

chouseknecht:
- Make shipit command read and parse container.yml using the same methods as other commands.
- Remove reliance on docker/compose module for reading/parsing
- Add the --save-config option
- Make shipit work with all container engines


j00bar:
Plz update shipit.rst docs.

Issue #41, opened 2016-06-09
Unable to run ansible-container init


asemt:
Hey,

I get following errors running `ansible-container` (Git SHA id: `429dcc5`):

      $ ansible-container init
      Traceback (most recent call last):
        File "/home/asemt/coding/python/ansible-container-venv/bin/ansible-container", line 9, in <module>
          load_entry_point('ansible-container==0.1', 'console_scripts', 'ansible-container')()
        File "/home/asemt/coding/python/ansible-container-venv/local/lib/python2.7/site-packages/pkg_resources/__init__.py", line 558, in load_entry_point
          return get_distribution(dist).load_entry_point(group, name)
        File "/home/asemt/coding/python/ansible-container-venv/local/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2682, in load_entry_point
          return ep.load()
        File "/home/asemt/coding/python/ansible-container-venv/local/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2355, in load
          return self.resolve()
        File "/home/asemt/coding/python/ansible-container-venv/local/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2361, in resolve
          module = __import__(self.module_name, fromlist=['__name__'], level=0)
      ImportError: No module named container.cli

I followed [these instructions](https://github.com/ansible/ansible-container#to-install-ansible-container) to install `ansible-container` inside a virtualenv on a Ubuntu 14.04.4 LTS machine.

gaieges:
Ran into the same problem, [this fix](https://github.com/ansible/ansible-container/pull/42) seems to have worked for me

PR #42, opened 2016-06-09
fix "No module named container.cli" issue

gaieges:
Ran into issue #41 earlier, figured out what was going on and fixed it.  Not sure what the downstream impacts are given I'm not a python expert, will rely on reviews for guidance there.

Looks like there's no argument 'include' per the [find_packages doc](http://code.nabla.net/doc/setuptools/api/setuptools/setuptools.find_packages.html#setuptools.find_packages), so I removed it.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -8,7 +8,7 @@ reqs = [str(ir.req) for ir in install_reqs]
 setup(
     name='ansible-container',
     version=container.__version__,
-    packages=find_packages(include='container'),
+    packages=find_packages(),
     include_package_data=True,
     url='https://github.com/ansible/ansible-container',
     license='LGPLv3 (See LICENSE file for terms)',


chouseknecht:
I ran into this yesterday as well. The fix is to upgrade setup tools. The *include* option was added way back in version 3.3 according to the [project change log](https://github.com/pypa/setuptools/blob/a73ca4fdcd1ba386159eb6ceb66d40af674a034c/CHANGES.rst)

I reinstalled setup tools using instructions [found here](https://github.com/pypa/setuptools). That got me to the latest release and fixed the issue.

gaieges:
Hmm .. Are you sure that's what's going on?  I'm using a stock xenial vm with a more recent version of setuptools than 3.3:

```bash
$ python -c 'import setuptools; print setuptools.__version__'
20.7.0
```

After upgrading to 23 in the doc you provided above, I seem to get a very similar thing:

```bash
$ python -c 'import setuptools; print setuptools.__version__'
23.0.0

$ python setup.py install
## .. snip ..
Using /usr/local/lib/python2.7/dist-packages/setuptools-23.0.0-py2.7.egg
Finished processing dependencies for ansible-container==0.1

$ ansible-container init
  File "/usr/local/bin/ansible-container", line 9, in <module>
    load_entry_point('ansible-container==0.1', 'console_scripts', 'ansible-container')()
  File "build/bdist.linux-x86_64/egg/pkg_resources/__init__.py", line 542, in load_entry_point
  File "build/bdist.linux-x86_64/egg/pkg_resources/__init__.py", line 2569, in load_entry_point
  File "build/bdist.linux-x86_64/egg/pkg_resources/__init__.py", line 2229, in load
  File "build/bdist.linux-x86_64/egg/pkg_resources/__init__.py", line 2235, in resolve
ImportError: No module named container.cli
```

Good catch on the 'include', couldn't find it elsewhere in the docs.

Open to doing something wrong here.. lmk what you think.

chouseknecht:
Ah. Figured it out. It should be `find_packages(include='container.*')`. Fixing.

Issue #44, opened 2016-06-10
Docker cli in orchestration container depends on, but doesn't include shared libs


gaieges:
Getting the following exception when trying to gather facts in the first ansible play within the ''ansible_container'' orchestration container:

```
ansible-container_1  | Using /ansible-container/ansible/ansible.cfg as config file
ansible-container_1  | Loaded callback default of type stdout, v2.0
ansible-container_1  |
ansible-container_1  | PLAYBOOK: main.yml *************************************************************
ansible-container_1  | 2 plays in main.yml
ansible-container_1  |
ansible-container_1  | PLAY [host1] ********************************************************************
ansible-container_1  |
ansible-container_1  | TASK [setup] *******************************************************************
ansible-container_1  | /usr/bin/docker: error while loading shared libraries: libapparmor.so.1: cannot open shared object file: No such file or directory
ansible-container_1  | An exception occurred during task execution. The full traceback is:
ansible-container_1  | Traceback (most recent call last):
ansible-container_1  |   File "/src/ansible/lib/ansible/executor/task_executor.py", line 124, in run
ansible-container_1  |     res = self._execute()
ansible-container_1  |   File "/src/ansible/lib/ansible/executor/task_executor.py", line 411, in _execute
ansible-container_1  |     self._connection = self._get_connection(variables=variables, templar=templar)
ansible-container_1  |   File "/src/ansible/lib/ansible/executor/task_executor.py", line 636, in _get_connection
ansible-container_1  |     connection = self._shared_loader_obj.connection_loader.get(conn_type, self._play_context, self._new_stdin)
ansible-container_1  |   File "/src/ansible/lib/ansible/plugins/__init__.py", line 349, in get
ansible-container_1  |     obj = obj(*args, **kwargs)
ansible-container_1  |   File "/src/ansible/lib/ansible/plugins/connection/docker.py", line 77, in __init__
ansible-container_1  |     docker_version = self._get_docker_version()
ansible-container_1  |   File "/src/ansible/lib/ansible/plugins/connection/docker.py", line 116, in _get_docker_version
ansible-container_1  |     cmd_output = subprocess.check_output(cmd)
ansible-container_1  |   File "/usr/local/lib/python2.7/subprocess.py", line 573, in check_output
ansible-container_1  |     raise CalledProcessError(retcode, cmd, output=output)
ansible-container_1  | CalledProcessError: Command '['/usr/bin/docker', 'version']' returned non-zero exit status 127
ansible-container_1  |
[...]

PR #45, opened 2016-06-11
Install docker in the ansible build container

chouseknecht:
This removes the volume bind of the host's docker executable to the build container. We hit multiple examples where the host's docker daemon is using a storage driver not compatible with the Python 2.7 (Debian jessie) base image container. Installing docker make the image build a little bit slower, but it's the only way to insure docker works correctly.


chouseknecht:
Fixes #44

PR #50, opened 2016-06-12
Add missing https url-scheme to the upstream git repo-url

dictvm:
Trying to clone the repo without an url scheme leads to an error, because git defaults to ssh: 

```
git clone git@github.com:ansible/ansible-container.git
Cloning into 'ansible-container'...
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

It works however, when you try to clone via https instead of ssh.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ Prerequisites:
 
 Clone the repo:
 
-    $ git clone git@github.com:ansible/ansible-container.git
+    $ git clone https://github.com/dictvm/ansible-container.git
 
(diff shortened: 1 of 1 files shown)


j00bar:
Modify your patch to keep the repo as this one, not your fork, and I'll merge. Thanks for catching this!!

dictvm:
I missed that, sorry. Fixed with f11bbf96c4519386d3d1bb6b62758a16ec587143.

PR #55, opened 2016-06-15
Add --repository to push subcommand.

chouseknecht:
Allow the user to set a custom repository name. Defaults to registry username. Needed in the case of the GKE where the repository name must be the project ID.


iainlbc:
Will this land in a release soon? Seems odd it was merged to master on Jun 14 but not in 0.2.0 that dropped a few weeks back.

Ran into this when attempting to push aws ecr 

Thanks!

PR #62, opened 2016-06-17
Use nginx in init command, uncomment the example

cdrage:
Uncomment the example and use NGINX

Rather than use Ubuntu and Apache (resulting in a larger image to
download), use nginx which has larger community support within the
Docker eco-system.

I also uncomment the example so people can use ansible-container
straight away, rather than having to uncomment each line.

### DIFF
--- a/container/templates/ansible/container.j2.yml
+++ b/container/templates/ansible/container.j2.yml
@@ -2,12 +2,9 @@ version: "1"
 services:
   # Add your containers here, specifying the base image you want to build from
   # For example:
-  #
-  # web:
-  #   image: ubuntu:trusty
-  #   ports:
-  #     - "80:80"
-  #   command: ['/usr/sbin/dumb-init', '/usr/bin/apache2ctl', '-D', 'FOREGROUND']
-  #   dev_overrides:
-  #     environment:
-  #       - "DEBUG=1"
+  web:
+    image: nginx
+    ports:
+      - "80:80"
+    environment:
+      - NGINX_PORT=80


cdrage:
I realize that there is a playbook generated as well, perhaps we can do the same to the `main.yml` that is generated too?

geerlingguy:
I would think the example should still be commented, though—there are plenty of use cases where someone wouldn't want a `web` image, and it's safer to not create one explicitly (e.g. opt-in and fail if the user hasn't set things up properly rather than opt-out).

gaieges:
My 2c: I agree with @cdrage that having a init command with a working example is a great way to get users up and running quickly.  

As a compromise between the two ideas, how about a vagrant init like approach where you specify some sort of named template config on the command line that outputs a working example?

cdrage:
@geerlingguy @gaieges 

After testing ansible-container a bit more, I see why we use ubuntu + apache (we want to show the approach of using main.yml to configure everything, etc. considering the nginx container (i think) is using alpine)

So I'm leaning back to using the Ubuntu example specified. Although maybe with Nginx instead of Apache + uncommented? :).

gaieges:
Seems reasonable to start with the smallest image possible.  Not sure if that's ubuntu or debian or something else.  

Whether it's apache or nginx isn't too big of a deal I think, they're both really well known these days.

gregdek:
So after having gone through this example a few times myself, I think I'm a -1 to having the example code uncommented.

cdrage:
I'm going to close this for now and bring up a different PR once we can agree on what within #65 :)

PR #85, opened 2016-06-23
Add Makefile

cdrage:


cdrage:
Also adds methods to clean, install and test locally

Makefile
```diff
@@ -0,0 +1,14 @@
...
+	$(PYTHON) setup.py install
+
+.PHONY: clean
+clean:
+	$(PYTHON) setup.py clean --all
+
+.PHONY: test
+test:
+	pip install -r requirements.txt -r test/requirements.txt
+	$(PYTHON) -m pytest -vv
```
mattclay: To get the most out of the tests (ex: code coverage), use the shell scripts in ```test/utils``` to run the tests.

Makefile
```diff
@@ -0,0 +1,14 @@
...
+install:
+	$(PYTHON) setup.py install
+
+.PHONY: clean
+clean:
+	$(PYTHON) setup.py clean --all
+
+.PHONY: test
+test:
+	pip install -r requirements.txt -r test/requirements.txt
```
mattclay: Installing requirements should probably be done with a separate make target, so it doesn't need to be executed with each test run.

cdrage:
Should we modify tests to run similar to ansible then? https://github.com/ansible/ansible/blob/devel/Makefile#L103 @mattclay instead of using the bash script ^^

mattclay:
@cdrage The ansible tests only provide code coverage for unit tests. Both unit and integration tests for ansible-container have code coverage, but doing so relies on using bash scripts.

What advantage is there in using make instead of bash scripts in this case?

cdrage:
@mattclay Centralized area where people can look at the make file and go "Oh hey! That's how I run the tests" instead of guessing. Similar to how Ansible has it, numerous other projects, etc.

Helps organize things a little :)

mattclay:
@cdrage It makes sense to have a test target in the Makefile, as you stated. I don't think there's a need to eliminate the bash scripts though.

Makefile
```diff
@@ -0,0 +1,14 @@
+PYTHON ?= /usr/bin/python
```
mattclay: Use ```python``` instead of specifying a full path. That will make it work with alternative paths, such as when using a python virtual environment.

Makefile
```diff
@@ -0,0 +1,14 @@
...
+DOCKER ?= /usr/bin/docker
```
mattclay: DOCKER is currently unused.

cdrage:
@mattclay updated

PR #129, opened 2016-07-22
Prebaked builder container

j00bar:


container/templates/builder.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+test '(! -f /src/requirements.txt)' || pip install --no-cache-dir -q -U -r /src/requirements.txt
+$*
```
gregdek: In quotes? "$*" ? Is this why it's failing Travis? Seems pedantic, but...

Issue #91, opened 2016-06-24
Support detached run, stop, and restart


milky-milk:
Hello.
Is it possible to use something like `docker-compose -d` or `docker run -d image` with ansible-container? 
I have tried to run a container with nginx with ansible-container, but it exits with `exit 0`
```
version: "1" 
services:
  nginx:
    image: centos:7
    ports:
      - "80:80" 
    command: ['/usr/bin/dumb-init', '/usr/sbin/nginx', '-c', '/etc/nginx/nginx.conf' ]
    volumes:
      - /var/www:/usr/share/nginx/html
```
```
Attaching to ansible_ansible-container_1
Cleaning up Ansible Container builder...
Attaching to ansible_nginx_1
ansible_nginx_1 exited with code 0
```

j00bar:
We can certainly support detached mode. We would have to also implement stop/restart.

sebastienbonami:
+1

chouseknecht:
Completed by @cdrage.

containscafeine:
@chouseknecht s/cdrage/containscafeine? :P

PR #136, opened 2016-07-27
implement detached mode for run, fix #131

### LINKED ISSUE #131: Ability to run applications in detached mode
Hi,

It would be great if I could run `ansible-container run` in detached mode (with a `-d`, maybe?) so I can get my current terminal back.

Does it make sense?

containscafeine:
##### ISSUE TYPE
 - Feature Pull Request

##### SUMMARY
Issue #91 and #131 talked about the need of deploying the application in detached mode, which is what I have attempted to do in this PR.

Since `docker.engine.orchestrate` calls `get_config_for_...`, which in turn calls `orchestrate` again, I had to remove `self.params['detached']` so that `get_config_for_...` does not run in detached mode; I set a `detached` flag to True, which I use in the end before `command.up` is executed to pass `-d` to `command.up`

```
Before:
# ansible-container run -d                       
usage: ansible-container [-h] [--debug] [--engine ENGINE_NAME]
                         [--project BASE_PATH]
                         {run,help,shipit,init,build,push} ...
ansible-container: error: unrecognized arguments: -d
```

```
After:
# ansible-container run -d
Cleaning up Ansible Container builder...
Deploying application in detached mode

# docker ps
CONTAINER ID        IMAGE                      COMMAND                  CREATED             STATUS              PORTS                          NAMES
ec0b6f5ab126        example-webclient:latest   "./tmp/client.sh"        11 minutes ago      Up 10 minutes                                      ansible_webclient_1
33b1acd02827        example-webserver:latest   "/usr/sbin/apachectl "   11 minutes ago      Up 10 minutes       80/tcp, 0.0.0.0:80->8080/tcp   ansible_webserver_1
```

### DIFF
--- a/container/cli.py
+++ b/container/cli.py
@@ -91,6 +91,9 @@ def subcmd_run_parser(parser, subparser):
     subparser.add_argument('--production', action='store_true',
                            help=u'Run the production configuration locally',
                            default=False, dest='production')
+    subparser.add_argument('-d', action='store_true',
+                           help=u'Deploy application in detached mode',
+                           dest='detached')
 
 def subcmd_help_parser(parser, subparser):
     return
--- a/container/cli.py
+++ b/container/cli.py
@@ -91,7 +91,7 @@ def subcmd_run_parser(parser, subparser):
     subparser.add_argument('--production', action='store_true',
                            help=u'Run the production configuration locally',
                            default=False, dest='production')
-    subparser.add_argument('-d', action='store_true',
+    subparser.add_argument('-d', '--detached', action='store_true',
                            help=u'Deploy application in detached mode',
                            dest='detached')
 
--- a/container/cli.py
+++ b/container/cli.py
@@ -52,6 +52,7 @@ AVAILABLE_COMMANDS = {'help': 'Display this help message',
                       'init': 'Initialize a new Ansible Container project',
                       'build': 'Build new images based on ansible/container.yml',
                       'run': 'Run and orchestrate built images based on container.yml',
+                      'stop': 'Stop the services defined in container.yml, if deployed',
                       'push': 'Push your built images to a Docker Hub compatible registry',
                       'shipit': 'Generate a deployment playbook to your cloud of choice.'}
 
@@ -94,6 +95,11 @@ def subcmd_run_parser(parser, subparser):
     subparser.add_argument('-d', '--detached', action='store_true',
                            help=u'Deploy application in detached mode',
                            dest='detached')
+
+def subcmd_stop_parser(parser, subparser):
+    subparser.add_argument('service', action='store',
+                           help=u'The specific services you want to stop',
+                           nargs='*')
 
 def subcmd_help_parser(parser, subparser):
     return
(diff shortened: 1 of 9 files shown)


chouseknecht:
Well, this works to start the application. But it does nothing to stop the application.

containscafeine:
Hey @chouseknecht, are you suggesting something like `ansible-container stop`?
I only fixed the running in detached mode with this PR(Issue #131), I would love to iterate on this to implement `stop` and `restart` too (Issue #91).
Would you want that in the same PR?
Hope I understood you correctly.

chouseknecht:
Adding `ansible-container stop` to this PR makes sense to me.

Rather than `restart`, I wonder if we should just surface the `--recreate` and `--no-recreate` options that are already part of `docker-compose up`. I think keeping the options the same will make it less confusing.

container/cli.py
```diff
@@ -90,6 +90,9 @@ def subcmd_run_parser(parser, subparser):
     subparser.add_argument('--production', action='store_true',
                            help=u'Run the production configuration locally',
                            default=False, dest='production')
+    subparser.add_argument('-d', action='store_true',
```
chouseknecht: Add '--detached' to be consistent with other ansible-container options.

container/docker/engine.py
```diff
@@ -317,6 +322,9 @@ def orchestrate(self, operation, temp_dir, hosts=[], context={}):
         command_options = self.DEFAULT_COMPOSE_UP_OPTIONS.copy()
         command_options[u'--no-build'] = True
         command_options[u'SERVICE'] = hosts
+        if 'detached' in locals():
```
j00bar: Why this route instead of `if self.params.get('detached')`?

container/cli.py
containscafeine: Done

container/docker/engine.py
containscafeine: @j00bar Since `self.get_config_for_run()` calls `self.hosts_touched_by_playbook()` which calls `self.orchestrate('listhosts' ...)` which fails if `self.params['detached']` is True, so I have deleted `self.params['detached']` and set `detached=True`, [here](https://github.com/containscafeine/ansible-container/blob/0e2477b684831c83f8a0b478c4e4aba65987e178/container/docker/engine.py#L281)

I have used `detached=True` [here](https://github.com/containscafeine/ansible-container/blob/0e2477b684831c83f8a0b478c4e4aba65987e178/container/docker/engine.py#L325)

containscafeine:
Hi @chouseknecht, did not get much time to work on this last week, picked this up today.

I have implemented -
- `ansible-container run -d`
- `ansible-container stop`
- `ansible-container stop <service>`

I have also added tests for the above changes.
Removing [WIP], this is up for review again.

Also, I'm +1 for `ansible-container restart`, imo it would be useful.

chouseknecht:
@containscafeine awesome! Thank you for working on this. Will review it this evening.

chouseknecht:
@containscafeine I agree. Makes sense to add *restart* and maybe a *--kill* option for the *stop* command.

containscafeine:
@chouseknecht on it! :)

PR #162, opened 2016-08-10
Feature/custom builder volumes

chouseknecht:
##### ISSUE TYPE
 - Feature Pull Request

##### SUMMARY
**Adds** the following to the *build* command:

```
 --with-volumes WITH_VOLUMES [WITH_VOLUMES ...], -v WITH_VOLUMES [WITH_VOLUMES ...]
                        Mount one or more volumes to the Ansible Builder
                        Container. Specify volumes using the Docker volume
                        format string. Separate multiple volume strings with
                        spaces.
  --with-variables WITH_VARIABLES [WITH_VARIABLES ...], -e WITH_VARIABLES [WITH_VARIABLES ...]
                        Define one or more environment variables in the
                        Ansible Builder Container. Format each variable as a
                        key=value string. Separate multiple variable strings
                        with spaces.
  --save-build-container
                        Leave the Ansible Builder Container intact upon build
                        completion. Use for debugging and testing.
```

**Makes** the tests runnable both locally and on Travis using a single command. Yes! To run tests:

```
$ /path/to/local/ansible-container/test/run.sh
```
Tests will execute inside of a container. If the local-test image does not already exist, the script will first build it using `ansible-container build`. It then runs the tests using `ansible-container run`.  Relies on the new --with-variables to make the magic work. The project for maintaining the image is found in test/local.

**Adds** integration tests for --with-variables and --with-volumes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ before_install:
   - sudo apt-add-repository 'deb http://archive.ubuntu.com/ubuntu trusty-backports universe'
   - sudo apt-get update -qq
   - sudo apt-get install -y shellcheck
-  - sudo apt-get install -y -o Dpkg::Options::="--force-confold" docker-engine
+  - sudo apt-get install -y -o Dpkg::Options::="--force-confold" docker-engine=1.11.2~trusty
 
 install:
   - pip install -r requirements.txt -r test/requirements.txt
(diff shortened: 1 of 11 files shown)


nonotest:
excellent additions!

PR #170, opened 2016-08-14
Don't mount basepath directly, put it in a data container instead

vrutkovs:
##### ISSUE TYPE
 - Feature Pull Request

##### SUMMARY

ansible-container mounts ```base_path``` directly, which might be dangerous. We should pack the tarball of ```ansible/``` folder in a container, export it as a volume and mount it in ```ansible-container```

This also enables using ansible-containers on remote docker hosts.

### DIFF
--- a/container/cli.py
+++ b/container/cli.py
@@ -89,6 +89,9 @@ def subcmd_build_parser(parser, subparser):
     subparser.add_argument('--save-build-container', action='store_true',
                            help=u'Leave the Ansible Builder Container intact upon build completion. '
                                 u'Use for debugging and testing.', default=False)
+    subparser.add_argument('--with-data-container', action='store_true',
+                           help=u'Put ansible code in a separate data container'
+                                u'Use for remote docker hosts.', default=False)
     subparser.add_argument('ansible_options', action='store',
                            help=u'Provide additional commandline arguments to '
                                 u'Ansible in executing your playbook. If you '
--- a/container/docker/engine.py
+++ b/container/docker/engine.py
@@ -12,6 +12,7 @@ import getpass
 import json
 import base64
 import pprint
+import datetime
 
 import docker
 from docker.client import errors as docker_errors
@@ -41,12 +42,17 @@ class Engine(BaseEngine):
     orchestrator_name = 'Docker Compose'
     builder_container_img_name = 'ansible-container'
     builder_container_img_tag = 'ansible-container-builder'
+    builder_data_container_name = 'ansible-data'
     default_registry_url = 'https://index.docker.io/v1/'
     default_registry_name = 'dockerhub'
     _client = None
     _orchestrated_hosts = None
     api_version = ''
     temp_dir = None
+
+    def __init__(self, *args, **kwargs):
+        super(Engine, self).__init__(*args, **kwargs)
+        self.builder_data_img_name = '%s-data' % self.project_name
 
     def all_hosts_in_orchestration(self):
         """
@@ -56,7 +62,16 @@ class Engine(BaseEngine):
         """
         return self.config.get('services', {}).keys()
 
-    def hosts_touched_by_playbook(self):
+    def cleanup_build_containers(self, with_data_container=False):
+        logger.info('Cleaning up Ansible Container builder...')
+        builder_container_id = self.get_builder_container_id()
+        self.remove_container_by_id(builder_container_id)
+        if with_data_container:
+            logger.info('Cleaning up Ansible Data Container builder...')
+            builder_data_container_id = self.get_builder_data_container_id()
+            self.remove_container_by_id(builder_data_container_id, force=True)
+
+    def hosts_touched_by_playbook(self, with_data_container=False):
         """
         List all hosts touched by the execution of the build playbook.
 
@@ -65,10 +80,9 @@ class Engine(BaseEngine):
         if not self._orchestrated_hosts:
             with teed_stdout() as stdout, make_temp_dir() as temp_dir:
                 self.orchestrate('listhosts', temp_dir,
-                                 hosts=[self.builder_container_img_name])
-                logger.info('Cleaning up Ansible Container builder...')
-                builder_container_id = self.get_builder_container_id()
-                self.remove_container_by_id(builder_container_id)
+                                 hosts=[self.builder_container_img_name],
+                                 context=dict(with_data_container=with_data_container))
+
                 # We need to cleverly extract the host names from the output...
                 logger.debug('--list-hosts\n%s', stdout.getvalue())
                 lines = stdout.getvalue().split('\r\n')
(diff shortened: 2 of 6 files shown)


chouseknecht:
@vrutkovs Thank you for contributing to ansible-container! We appreciate you taking the time to use ansible-container, and we definitely appreciate your willingness to jump in and help, so a big **thank-you** for putting your time and effort into this submission!

My initial reaction is this seems to adds a layer of complexity to the build process that feels unnecessary. I have a couple questions that may help me understand the use case...

First, how does putting the base path inside of a container automatically make it more safe? Whatever is on the base path will end up being touched or used by the build container. If the base path contains something *dangerous*, the build container will consume it regardless of whether a host mount or a volume_from mount is used.

Next, can you explain how this 'enables using ansible-container on remote docker hosts'?. Walk me through what the problem is and how this solves it. I'm not trying to be difficult. I want to understand the problem being solved.

If I'm thinking about this correctly (and maybe I'm not), you're trying to execute a build on a remote host, and that remote host is running a Docker daemon. The first step, I think, would be to get the project on the remote host, and that of course can be accomplished via your SCM (i.e. `git clone my-project`). The next step is to execute `ansible-container build` on the remote host. That could be done via an Ansible playbook using the command module. However, it would require having ansible-container and dependencies installed on the remote host. how does this PR help this process?

vrutkovs:
Thanks for a quick reply, here are my thoughts on why I started this PR:

>First, how does putting the base path inside of a container automatically make it more safe? Whatever is on the base path will end up being touched or used by the build container. If the base path contains something dangerous, the build container will consume it regardless of whether a host mount or a volume_from mount is used.

That's correct, however since we're mounting this folder in RW mode the "danger" can go both ways - ansible script may modify the host's filesystem, which is not a good idea. In this PR the ```ansible/``` folder will be put in a tarball and become a copy of the scripts and files to be used. 
These data containers will be tagged with the same timestamp as a resulting image and can be helpful to understand which code was used to build a specific container.

As a side effect this PR also enables NFS shares to be used as a code source (I don't think someone ever did that though) etc.

This is not the main point of this PR, however, the idea below was my the reason I started this:
>Next, can you explain how this 'enables using ansible-container on remote docker hosts'?. Walk me through what the problem is and how this solves it. I'm not trying to be difficult. I want to understand the problem being solved.

The problem: if we use Dockerfiles we are able to build a container on a remote docker host - all it takes is to set DOCKER_HOST env var and set up remote docker host. Currently, ansible-container doesn't support this: as we're mounting a base path inside ```ansible-container``` - so the code and the docker daemon should be running on the same machine.

>If I'm thinking about this correctly (and maybe I'm not), you're trying to execute a build on a remote host, and that remote host is running a Docker daemon

That's correct. Previously I was copying containers from dev to "prod" machine or replicating my environment on prod machine - rsyncing/git cloning the code, setting up ansible-container and running commands. The biggest problems I've encountered are:
a) storing secrets is rather complex - either the remote host should have necessary credentials
b) initially I selected Fedora Atomic, which is nicely tuned for containers, but getting git/ansible-container was rather complex. I had to fall back to Fedora Cloud instead.
c) In order to deploy I created a simple ```deploy.yml``` playbook - but I had to hardcode the repo. Also the gitlab I was using had self-signed certificates, so I also had to set options for git. In the end the whole deploy process looked weird and over-engineered.

Thanks to ```shipit kube/openshift``` things would be simplier, but I didn't feel like to setup Kubernetes/Openshift for an app with two containers.

>How does this PR make this process any easier?

In order to build and run the container on a remote host we only need to specify correct DOCKER_HOST and setup the remote docker daemon. No ansible-container, ansible or git are required on the host - all it needs is docker daemon. Now ansible-container was usable with Fedora/CentOS/RHEL Atomic.

Hope that answers most of your questions.

----
Note, that currently tests are failing (despite the status Travis reports) - I'm not entirely sure why, probably I didn't test 'ansible-container runs another ansible-container' case well enough.

chouseknecht:
@vrutkovs  I get the value in this, and it's actually clever. However, it's not something that everyone needs all the time. It makes sense when deploying to a remote host, but otherwise it's extra overhead. So with that in mind, let's make this an option rather than the default behavior. Let's add a command line option --with-data-container to the *build* command.

I have a fix for the test script to always surface the test container's exit code. I'll get that merged shortly. That's why travis is reporting success even though a test is broken. Also, while poking at tests discovered the --remove-orphans is missing from the *run* command. Investigating that and will fix as well.

vrutkovs:
>Let's add a command line option --with-data-container to the build command.

Yeah, agreed

chouseknecht:
@vrutkovs as far as the actual test error. It's coming from a failure of the hosts_touched_by_playbook method in docker/engine.py to return the correct set of hosts for the integration test *test_build_minimal_docker_container*. Looks like the inventory script ansible-container-inventory.py is returning the inventory from the *minimal* project when it should be returning the inventory for *minimal_sleep*. Hopefully that helps point you in the right direction.

j00bar:
Why are we under the impression that ansible-container's build process doesn't work with `DOCKER_HOST`? And note that using a volume would prevent `shipit` from writing out the roles to the `ansible/` directory outside of the build container.

vrutkovs:
>Why are we under the impression that ansible-container's build process doesn't work with DOCKER_HOST?

We're mounting base_path in ```build``` phase. If DOCKER_HOST has ```ansible/``` dir in the exact location it would work, otherwise it would be unable to start ```ansible-container```

>And note that using a volume would prevent shipit from writing out the roles to the ansible/ directory outside of the build container.

```--with-data-container``` should be used for ```build``` only, other commands are not affected as soon as we have the image built

j00bar:
Wouldn't a simpler approach be to replace `os.getcwd()` here with `os.path.curdir`?

vrutkovs:
```os.path.curdir``` will only return ```.``` on linux systems, in order to make it useful it should be wrapped in abspath - ```os.path.abspath(os.path.curdir)```

j00bar:
The reason I suggested `os.path.curdir` is that it would ultimately make its way into the Dockerfile saying you wanted the volume at `.` to be mounted at `/ansible-container/` - whereas `os.path.abspath(os.path.curdir)` or `os.getcwd()` would be evaluated on the host from which the `ansible-container` script was run. I'm actually not 100% sure how you specify the path to export into a volume in a docker-compose file when the Docker engine is remote.

vrutkovs:
>I'm actually not 100% sure how you specify the path to export into a volume in a docker-compose file when the Docker engine is remote.

Right, there is no reliable way to specify the path in ```docker-compose.yml```, keeping in mind that docker host can be remote. This is why we should take files in ```base_path```, put them in a container, send it to a (remote) docker host and mount this container as a volume.

This way we'll make sure that:
a) all files can be passed to a remote host, as we only know docker host and port, but no other details.
b) no other files except the ones in ```ansible/``` directory can be used
c) we can track which ansible files were used for a specific version of the container (data containers will be tagged same way as resulting container)

j00bar:
Can you help me understand the use case? I feel like a remote Docker host is an uncommon setup, especially in a developer's or CI environment.

vrutkovs:
I'm not sure how common is it. I'd say this feature will be eventually requested again, as people might use:

a) remote host with Fedora/CentOS/RHEL Atomic (it has no git/pip, so fetching code and installing ansible-container is not an option)
b) a remote machine with docker, which is much more powerful than their laptop/desktops - and this remote host disallows mounting folders from developer machines.

j00bar:
I'm less familiar with Atomic as I am Docker Machine. I know Docker Machine exports into the filesystem the "users" directory (/home, /Users/, C:\Users), and that way you can still reference filesystems to export into containers run by Docker Machine. Does Atomic not have a similar export?

vrutkovs:
No, its not. Its more of a production class system, see http://www.projectatomic.io/docs/introduction/ for more information

j00bar:
Looking at the developer bundle, it seems like it does do that same export?

https://github.com/projectatomic/adb-atomic-developer-bundle/blob/master/components/centos/centos-k8s-singlenode-setup/Vagrantfile#L57-L62

Is that the wrong sort of place to be looking?

vrutkovs:
Its a Vagrant configuration, it has nothing to do with Atomic itself. I'm usually running my machines in AWS or private Openstack instance. Setting up sshfs on server to fetch several files is an overkill IMHO

j00bar:
Given that we presently work out of the box with the Atomic Developer Bundle, Docker, and Docker Machine, with the same caveat in each that you're working out of `$HOME`, which strikes me as the dominant use case, I'm unconvinced about this being more than an edge case specific to your chosen workflow, and you've yourself already identified a workaround with network filesystems like sshfs. If other members of the wider community feel strongly otherwise, I'm happy to reconsider. But for the time being, this is WONTFIX.

vrutkovs:
This PR has nothing to do with Atomic Developer Bundle and Docker Machine. And no, it doesn't work out of the box with a remote Docker daemon. 

>you're working out of $HOME, which strikes me as the dominant use case

So, ansible-container is designed to be a developer toy? Why would you oppose a change, which would help to be good for stage or prod servers, where "working out of ```$HOME```" is a no-no?

>you've yourself already identified a workaround with network filesystems like sshfs

This is a horrible solution, its good enough for a developer machine, not a test/stage or a prod server.

I've explained numerous times in the comments the reason for the change (which doesn't affect default behaviour) and the benefits of this approach.

j00bar:
I hear and understand your frustration and disappointment at my decision. Ansible Container is not designed to be a developer toy. It's designed to enable a container workflow. The `build` in this workflow would be executed on the developer's workstation and on a CI server. Docker Machine out of the box sets `DOCKER_HOST` and connects to the Docker engine over a secured TCP socket. This works because the Virtualbox instance exports `/home`/`/Users`/`C:/Users` into the boot2docker machine. Looking at the Atomic Developer Bundle, which presumably is the the Atomic-blessed parallel to Docker Machine, exports `$HOME` into the kvm instance. A CI server would arguably spin up a worker running its own Docker engine for testing.

Our solution maintains parity with the flexibility of Docker Compose, which is fine by me. And as I've identified, your solution would still be a complicated workflow to get the shipit generated roles back out of your remote host.

We're a new project and we're sussing out the directions the community wishes to see us go. The feedback we've gotten so far is that the local Docker daemon, local Docker Machine, and local Atomic Host cover the vast majority of use cases. Running Atomic Host on an EC2 instance or a private OpenStack instance I'd classify as a minority case. As every feature has a cost in code complexity, I'm endeavoring to follow an 80/20 rule here.

Like I said, I'm more than happy to be overridden if there's demand in the broader community to go this direction. But for the moment, that's my inclination. I understand it leaves Ansible Container non-ideal for your workflow, but I hope you'll continue contributing to the project anyway. Thank you again!

Issue #120, opened 2016-07-10
Support for Variables in Container.yml


darkn3rd:
### DESCRIPTION

Docker-Compose has support to source in environment variables from the current running environment.  It can also default them in a `./.env`.  I would also like the source them as well, or specify them in the file.  It would be nice to have shared variables both in `container.yml` and `main.yml`.

Example:

```yaml
version: '1'
vars:
  web_port: 8080
  app_root: "/usr/src/app"
services:
  express:
    image: node:argon
    command: [ "npm", "start" ]
    working_dir: "{{ app_root }}"
    environment:
       EXPRESS_PORT: "{{ web_port }}"
    ports:
      - "{{ web_port }}:{{ web_port }}"
registries: {}
```

dmsimard:
Yeah... :+1: this would be nice.

chouseknecht:
While working on moving Galaxy to containers and Galaxy deployment to using Ansible Container, I ran into this as well. We can fix the lack of .env support by adjusting the base_path value we pass into project_from_options() [here](https://github.com/ansible/ansible-container/blob/develop/container/docker/engine.py#L335). If we passed in 'self.base_path/ansible' it would automatically pick up '.env'. However, it would change the value of ${PWD} in container.yml. 

Changing the value of ${PWD} will break existing container.yml files, if they container the env_file directive like so:

```
services:
   foo:
      env_file:
         - ${PWD}/ansible/galaxy/environment.var
```

Any strong objections to making this change?

As to the use of .env... IMHO, using .env is awful. I don't want variables in a hidden file. Things should be obvious and self documenting. Using a hidden file make things not obvious and not self documenting. I would like to have a vars file much the same as ansible-playbook has *--extra-vars* and have the ability to pass in a vars file to template container.yml as described above. We'll have to document order of precedence, but overall should not be too difficult. I think adding this feature gives yet another advantage over using Dockerfile and straight-up Docker compose. Am I wrong in thinking this way?

dmsimard:
Vars should just be pulled from Ansible ? group_vars, host_vars, defined in inventory or role defaults and so on.

chouseknecht:
I don't think we want to repeat what Ansible already does or duplicate Ansible's code. Nor do we want to blend ansible-playbook with ansible-container. They are separate things and should remain separate.

My point was that we should be able to pass in a *basic* yaml variable file to the build command. Maybe something like... `ansible-container build --vars-file vars.yml`. And vars.yml might contain...

```
[...]

Issue #108, opened 2016-07-05
Reference for Container.yml


darkn3rd:
Are there any documentation reference for the tokens that can be used in the container reference?  I'm not sure how to convert my `Dockerfile` into a `container.yml`.   The current reference documentation just documents that commands.

mablae:
Hello @darkn3rd , 

you do not directly convert your `Dockerfile` to `container.yml`. 

The way doing this would be `docker build` to create an Container Image from the `Dockerfile` and refer to that "Container Image" in the `container.yml`  which is almost equal to the `docker-compose.yml` format. Then you would run `ansible-container` to provision a (new) container with this image as a base plus your `main.yml` playbook applied.

darkn3rd:
I was actually finally able to convert my `Dockerfile` to `container.yml` and `main.yml`.  I had to move functionality that I specified in the `Dockerfile` to `container.yml`, and specific package install and other configurations moved to `main.yml`.

I had some initial confusion as the docs are sparse, ran into bugs (expected, but caused the most problems).  As a newbie with basic Ansible, Docker Engine, and Docker-Compose knowledge, it was rough to get started.  I cannot imagine someone missing such pre-reqs.   Also, I was thinking it would be nice to have mini-tutorials like Docker-Compose has with Django, Flask, Rails, WordPress, which would help getting off the ground.

j00bar:
The docs do definitely need continual upkeep and love, and the `ansible-container-examples` repo needs some additional case studies. Contributions welcome until House or I can get to penning some on our own.

PR #181, opened 2016-08-19
Support jinja2 styles variables in container.yml

chouseknecht:
##### ISSUE TYPE
Feature Pull Request
 
##### SUMMARY
Per issue #120 adds support for jinja2 style variables in container.yml.

Modifies AnsibleContainerConfig to handle template expansion. AnsibleContainerConfig resolves variable values and performs the expansion on initialization. It uses the following order of precedence when resolving variables: 

- environment vars (any environment vars starting with AC)
- variables provided in files using new *--var-files* option
- and finally a new 'defaults' section added to the top level of container.yml.

For example, given the following:

- environment var AC_DEBUG=1
- devel.yml as a vars file that defines debug=0
- defaults section in container.yml defining debug: 2

The environment variable would win, giving *debug* a value of 1.

Here is an example container.yml taken from the new test project:

```
version: "1"
services:
   web:
     image: {{ web_image }}
     ports: {{ web_ports }}
     command: ['sleep', '10']
     dev_overrides:
       environment:
         - DEBUG={{ debug }}

   db: {{ db_service }}

registries: {}
```
Variables can be raw as above or enclosed in quotes. If enclosed in quotes, the result after template expansion will remain enclosed in quotes. Primitives and complex data structures can be substituted. 

Here's the var file used in the test:

```
---
debug: 1
web_ports:
  - 8000:8000
web_image: "python:2.7"
db_service:
  image: "python:2.7"
  command: sleep 10
  expose:
    - 5432
  environment:
    POSTGRES_DB_NAME: foobar
    POSTGRES_USER: admin
    POSTGRES_PASSWORD: admin
```
To build the project:

```
ansible-container --var-files=devel.yaml build --local-build
```

To run the project

```
ansible-container --var-files=devel.yaml run
```

To push the project:

```
ansible-container --var-files=devel.yaml push
```

To ship the project:

```
ansible-container --var-files=devel.yaml shipit
```

If using variables in container.yml, a mechanism for resolving variable values must be supplied with each command, as each command attempts to access container.yml. The mechanism can be --var-files (as above), or it can be the defaults section in the container.yml, environment vars, or a combination of all of the above. In other words, users will not necessarily always have to provide --var-files.

Template expansion occurs prior to the start of the build. It occurs just before the initial containers are created for each service. The containers are instantiated. The build process then runs the playbook, which does not rely on knowing the configuration of each service. It only requires service names. Any post build steps happen on the host where template expansion has already occurred. 

All that being said, the inventory script was updated. The update entails escaping any raw variables to allow them to pass PyYaml parsing. To work around this for testing I employed the --local-build option, which enables using the modified script. Once this PR is merged we will need to update the build image on Docker Hub.

This PR includes unit and integration tests. So there is at least an example, however it does not include documentation. After this is merged, I will take the responsibility of immediately updating docs and creating an example (outside of the new test project) using variable. The docs will be added as part of resolving #108 *Reference for containery.yml*, which is assigned to me.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,21 +1,24 @@
 Ansible Container has been contribued to by the following authors:
 (This list is automatically generated - please file an issue for corrections)
 
-Chris Houseknecht <chouseknecht@gmail.com>
+Chris Houseknecht <chouseknecht@ansible.com>
 Joshua "jag" Ginsberg <jag@flowtheory.net>
 Matt Clay <matt@mystile.com>
+Shubham Minglani <shubham@linux.com>
 Greg DeKoenigsberg <greg.dekoenigsberg@gmail.com>
 Sandra Wills <swills@ansible.com>
 Charlie Drage <charlie@charliedrage.com>
+chouseknecht <chousek@redhat.com>
 Will Thames <wthames@redhat.com>
 Todd Barr <tbarr@ansible.com>
+Arnaud Moret <arnaud.moret@gmail.com>
 Daniel Heitmann <dictvm@horrendum.de>
-Evan Zeimet <evan.zeimet@cdw.com>
 Abhishek Pratap Singh <abhishek@linux.com>
+Evan Zeimet <evan.zeimet@cdw.com>
+Gerard Braad <me@gbraad.nl>
 Sidharth Surana <ssurana@vmware.com>
-Ali Asad Lotia <ali.asad.lotia@gmail.com>
 Andrea De Pirro <andrea.depirro@yameveo.com>
+Ali Asad Lotia <ali.asad.lotia@gmail.com>
 Chrrrles Paul <chrrrles@users.noreply.github.com>
 Jeff Geerling <geerlingguy@mac.com>
 Shea Stewart <shea.stewart@arctiq.ca>
-Shubham Minglani <shubham@linux.com>
--- a/container/docker/engine.py
+++ b/container/docker/engine.py
@@ -491,7 +491,8 @@ class Engine(BaseEngine):
                     dict(
                         user='root',
                         working_dir='/',
-                        command='sh -c "while true; do sleep 1; done"'
+                        command='sh -c "while true; do sleep 1; done"',
+                        entrypoint=[]
                     )
                 )
             if not self.params['rebuild']:
@@ -538,7 +539,8 @@ class Engine(BaseEngine):
                 dict(
                     user='root',
                     working_dir='/',
-                    command='sh -c "while true; do sleep 1; done"'
+                    command='sh -c "while true; do sleep 1; done"',
+                    entrypoint=[]
                 )
             )
         return compose_config


container/cli.py
```diff
@@ -171,6 +172,10 @@ def commandline():
     parser.add_argument('--project', '-p', action='store', dest='base_path',
                         help=u'Specify a path to your project. Defaults to '
                              u'current working directory.', default=os.getcwd())
+    parser.add_argument('--var-files', action='store',
```
j00bar: How would multiple files work with order of inheritance? I'm going to suggest we start with just one.

container/config.py
```diff
@@ -17,34 +20,113 @@ class AnsibleContainerConfig(Mapping):
...
+        '''
         try:
             ifs = open(self.config_path, 'r')
         except Exception:
             raise AnsibleContainerConfigException("Failed to open %s. Are you in the correct directory?" %
                                                   self.config_path)
         try:
-            config = yaml.safe_load(ifs)
+            config = ifs.read()
+            config = re.sub(r"}}(?!\"|\')", "}}'", re.sub(r"(?<!\"|\'){{", "'{{", config, flags=re.M), flags=re.M)
```
j00bar: Plz comment regex's with intent so people don't have to decipher them.

container/config.py
```diff
@@ -17,34 +20,113 @@ class AnsibleContainerConfig(Mapping):
...
+        logger.debug('Getting AC environment variables...')
+        new_vars = {}
+        for var, value in os.environ.iteritems():
+            matches = re.match(r'^AC_(.+)$', var)
+            if matches:
+                new_vars[matches.group(1).lower()] = value
+        return new_vars
+
+    def _get_variables_from_file(self, file):
+        file_path = os.path.expandvars(os.path.expanduser(file))
```
j00bar: Consider `normpath` or `abspath` versus these.

container/config.py
```diff
@@ -17,34 +20,113 @@ class AnsibleContainerConfig(Mapping):
...
+        if not os.path.isfile(file_path):
```
j00bar: Plz comment with intent here.

container/config.py
```diff
@@ -17,34 +20,113 @@ class AnsibleContainerConfig(Mapping):
...
+            if not os.path.isfile(file_path):
+                file_path = os.path.join(self.base_path, 'ansible', file)
+        if not os.path.isfile(file_path):
+            raise AnsibleContainerConfigException("Unable to locate %s. Check that the file exists and you"
+                                                  "have read access." % file_path)
+        try:
+            fs = open(file_path, 'r')
+            data = fs.read()
+            fs.close()
+        except Exception:
```
j00bar: Limit this to `OSError` please.

container/config.py
```diff
@@ -17,34 +20,113 @@ class AnsibleContainerConfig(Mapping):
...
+            raise AnsibleContainerConfigException("Failed to open %s. Check that the file exists and you "
+                                                  "have read access." % file_path)
+        try:
```
j00bar: File extension should be sufficient to determine type.

container/config.py
```diff
@@ -17,34 +20,113 @@ class AnsibleContainerConfig(Mapping):
...
+            config = json.loads(data)
+        except:
+            # Failed to load as JSON. Let's try YAML.
+            try:
+                config = yaml.safe_load(data)
+            except yaml.YAMLError as exc:
+                raise AnsibleContainerConfigException("Reading %s\n. YAML exception: %s" % (file_path, str(exc)))
```
j00bar: Not unicode safe.

container/config.py
```diff
@@ -17,34 +20,113 @@ class AnsibleContainerConfig(Mapping):
...
+        return new_vars
+
+    def _get_variables_from_file(self, file):
+        file_path = os.path.expandvars(os.path.expanduser(file))
+        if not os.path.isfile(file_path):
+            file_path = os.path.join(self.base_path, file)
+            if not os.path.isfile(file_path):
+                file_path = os.path.join(self.base_path, 'ansible', file)
+        if not os.path.isfile(file_path):
+            raise AnsibleContainerConfigException("Unable to locate %s. Check that the file exists and you"
```
j00bar: Not unicode safe.

container/templates/ansible-container-inventory.py
```diff
@@ -4,10 +4,13 @@
 import json
 import sys
 import argparse
+import re
 
 def config_keys():
     ifs = open('/ansible-container/ansible/container.yml', 'r')
-    config = yaml.safe_load(ifs)
+    config = ifs.read()
+    config = re.sub(r"}}(?!\"|\')", "}}'", re.sub(r"(?<!\"|\'){{", "'{{", config, flags=re.M), flags=re.M)
```
j00bar: Plz document regex. It seems we probably want to go ahead and re-load/process config, no?

PR #183, opened 2016-08-23
Jinja2 template rendering in container.yml

chouseknecht:
##### ISSUE TYPE
Feature Pull Request
 
##### SUMMARY
Per issue #120 adds support for jinja2 style variables in container.yml.

Modifies AnsibleContainerConfig to support Jinja2 template rendering. AnsibleContainerConfig resolves variable values and performs the expansion on initialization. It uses the following order of precedence when resolving variables: 

- environment vars (any environment vars starting with AC)
- variables provided in a file using new *--var-file* option
- and finally a new 'defaults' section added to the top level of container.yml.

For example, given the following:

- environment var AC_DEBUG=1
- devel.yml as a var file that defines debug=0
- defaults section in container.yml defining debug: 2

The environment variable would win, giving *debug* a value of 1.

Here is an example container.yml taken from the new test project:

```
version: "1"
services:
   web:
     image: {{ web_image }}
     ports: {{ web_ports }}
     command: ['sleep', '10']
     dev_overrides:
       environment:
         - DEBUG={{ debug }}

   db: {{ db_service }}

registries: {}
```
Variables can be raw as above or enclosed in quotes. If enclosed in quotes, the result after template expansion will remain enclosed in quotes. Primitives and complex data structures can be substituted. 

Here's the var file used in the test:

```
---
debug: 1
web_ports:
  - 8000:8000
web_image: "python:2.7"
db_service:
  image: "python:2.7"
  command: sleep 10
  expose:
    - 5432
  environment:
    POSTGRES_DB_NAME: foobar
    POSTGRES_USER: admin
    POSTGRES_PASSWORD: admin
```
To build the project:

```
ansible-container --var-file devel.yaml build --local-build
```

To run the project

```
ansible-container --var-file devel.yaml run
```

To push the project:

```
ansible-container --var-file devel.yaml push
```

To ship the project:

```
ansible-container --var-file devel.yaml shipit
```

If using variables in container.yml, a mechanism for resolving variable values must be supplied with each command, as each command attempts to access container.yml. The mechanism can be --var-file (as above), or it can be the defaults section in the container.yml, environment vars, or a combination of all of the above. In other words, users will not necessarily always have to provide --var-file.

Template expansion occurs prior to the start of the build. It occurs just before the initial containers are created for each service. The containers are instantiated. The build process then runs the playbook, which does not rely on knowing the configuration of each service. It only requires service names. Any post build steps happen on the host where template expansion has already occurred. 

All that being said, the inventory script was updated. The update entails escaping any raw variables to allow them to pass PyYaml parsing. To work around this for testing I employed the --local-build option, which enables using the modified script. Once this PR is merged we will need to update the build image on Docker Hub.

This PR includes unit and integration tests. So there is at least an example, however it does not include documentation. After this is merged, I will take the responsibility of immediately updating docs and creating an example (outside of the new test project) using variable. The docs will be added as part of resolving #108 *Reference for containery.yml*, which is assigned to me.

### DIFF
--- a/container/cli.py
+++ b/container/cli.py
@@ -97,6 +97,7 @@ def subcmd_build_parser(parser, subparser):
                                 u'prefix your extra options. Use this feature with '
                                 u'caution.', default=u'', nargs='*')
 
+
 def subcmd_run_parser(parser, subparser):
     subparser.add_argument('service', action='store',
                            help=u'The specific services you want to run',
@@ -171,6 +172,10 @@ def commandline():
     parser.add_argument('--project', '-p', action='store', dest='base_path',
                         help=u'Specify a path to your project. Defaults to '
                              u'current working directory.', default=os.getcwd())
+    parser.add_argument('--var-files', action='store',
+                        help=u'One or more YAML or JSON formatted files providing variables for '
+                             u'template expansion in container.yml.', default=None, nargs='*')
+
     subparsers = parser.add_subparsers(title='subcommand', dest='subcommand')
     for subcommand in AVAILABLE_COMMANDS:
         logger.debug('Registering subcommand %s', subcommand)
--- a/container/cli.py
+++ b/container/cli.py
@@ -172,9 +172,9 @@ def commandline():
     parser.add_argument('--project', '-p', action='store', dest='base_path',
                         help=u'Specify a path to your project. Defaults to '
                              u'current working directory.', default=os.getcwd())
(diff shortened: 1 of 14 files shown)


chouseknecht:
Moved this out of my *develop* branch to a feature branch. Hopefully we can merge this soon. In the meantime, need my develop branch back.

PR #184, opened 2016-08-23
Increment version to 0.2.0

chouseknecht:
##### ISSUE TYPE
 - Bugfix Pull Request

##### SUMMARY
Increments version to 0.2.0-pre. Should have done this as soon as 0.1.0 went out. Assuming we want to do another *beta* release (or whatever we're calling it) and not a 1.0.0 release.

### DIFF
--- a/container/__init__.py
+++ b/container/__init__.py
@@ -5,4 +5,4 @@ import logging
 
 logger = logging.getLogger(__name__)
 
-__version__ = '0.1.0'
+__version__ = '0.2.0-pre'


chouseknecht:
@j00bar incrementing version for 2 reasons:

- The version in *develop* after each release (which is now) will be the next release
- Updating docs and thought it might be nice to indicate the version for new features.

Build is failing because **Tag 0.2 not found in repository docker.io/ansible/ansible-container-builder**. 

We need to create a job in Jenkins that will allow us to update the ansible/ansible-container-builder images. I think I can do this. Will ping you when I think it's ready. Need credentials for pushing to Docker Hub.

chouseknecht:
Pushed a 2.0 tagged image to Docker Hub.

PR #198, opened 2016-08-29
refactor orchestrate, terminate, restart functions

containscafeine:
Put redundant code under `def bootstrap_env` so `orchestrate`, `terminate`, `restart` functions could reuse it.

### DIFF
--- a/container/docker/engine.py
+++ b/container/docker/engine.py
@@ -730,8 +730,9 @@ class Engine(BaseEngine):
     def bootstrap_env(self, temp_dir, behavior, operation, compose_option,
                       builder_img_id=None, context=None):
         """
-        Builds the environment for common operations for compose engine,
-        like orchestration, termination, etc.
+        Build common Docker Compose elements required to execute orchestrate,
+        terminate, restart, etc.
+        
         :param temp_dir: A temporary directory usable as workspace
         :param behavior: x in x_operation_extra_args
         :param operation: Operation to perform, like, build, run, listhosts, etc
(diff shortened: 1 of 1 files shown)


container/docker/engine.py
```diff
@@ -780,3 +726,60 @@ def print_version_info(self):
         client = self.get_client()
         pprint.pprint(client.info())
         pprint.pprint(client.version())
+
+    def bootstrap_env(self, temp_dir, behavior, operation, compose_option,
+                      builder_img_id=None, context=None):
```
chouseknecht: Set context={}, rather than context=None. Then lines 746 - 747 can go away.

container/docker/engine.py
```diff
@@ -780,3 +726,60 @@ def print_version_info(self):
...
+        """
+        Builds the environment for common operations for compose engine,
+        like orchestration, termination, etc.
```
chouseknecht: This sentence doesn't quite make sense. Should be something like: 

Build common Docker Compose elements required to execute orchestrate, terminate, restart, etc.

container/docker/engine.py
containscafeine: I had done that, but then my IDE threw some warning about using mutable default arguments, and then I arrived at [this](http://docs.python-guide.org/en/latest/writing/gotchas/), so I decided to go with `context=None`.
WDYT?

container/docker/engine.py
containscafeine: Fixed. Rebased.

Issue #168, opened 2016-08-11
help our users navigate SELinux


dustymabe:
##### ISSUE TYPE
 - Bug/Enhancement ?? Bug vs Enhancement depends on who you are.

##### OS / ENVIRONMENT
Fedora 24
```
$ ansible-container --debug version
Ansible Container, version 0.1.0
Linux, f24, 4.5.5-300.fc24.x86_64, #1 SMP Thu May 19 13:05:32 UTC 2016, x86_64
2.7.12 (default, Jul 18 2016, 10:55:51) 
[GCC 6.1.1 20160621 (Red Hat 6.1.1-3)] /bin/python
No Ansible Container project data found - do you need to run "ansible-container init"?
```

##### SUMMARY
Ansible-container mounts the current working directory into the guest. Currently we add `:Z` to the mount (one example is: https://github.com/ansible/ansible-container/blob/dc048196e2e4fbf6e605dece55003cd416c2472b/container/templates/build-docker-compose.j2.yml#L15). We need to help our users in different SELinux scenarios. Here are the cases:

1. SELinux enabled: User is on normal filesystem
2. SELinux enabled: User is on shared filesystem (nfs,sshfs,vboxfs)
3. SELinux disabled:
4. SELinux doesn't exist (Operating systems that don't support SELinux)

We should probably try to be smart and/or give good error messages here. Something like:

```
if: system is capable of selinux and selinux is enabled on this system
  if: the filesystem allow us to change the context of the files where we are;
    add :Z to volume mount
  else:
    PRINT WARNING statement saying things might not work
    don't add :Z
else:
  PRINT DEBUG statement
  don't add :Z
```

For fuse filesystems we can warn and say that users can set `sudo setsebool -P virt_sandbox_use_fusefs 0`. For nfs filesystems something like `setsebool -P virt_use_nfs on
&& setsebool -P virt_sandbox_use_nfs on`

dustymabe:
apparently the behavior where docker "errors" when selinux is not supported is actually not what is supposed to happen. 

After talking to @rhatdan I opened this bug: https://bugzilla.redhat.com/show_bug.cgi?id=1373610

When that gets fixed in Fedora (and elsewhere) then we actually shouldn't need to implement this.

dustymabe:
Turns out it has been decided that the current behavior is not a regression and is correct, so the BZ has been closed.

That means this does actually need to get implemented and won't be fixed in the lower level tools.

dalethestirling:
[...]

Issue #76, opened 2016-06-22
call files "build.yml" and "run.yml"?


gregdek:
Upon reflection, seems perhaps more intuitive to change "container.yml" to "run.yml", since that's the file we reference when running "ansible-container run", and to change "main.yml" to "build.yml", since that's the file we reference when running "ansible-container build".

Since we're throwing some new concepts at users, any assistance we can give them in mapping those concepts is useful.

j00bar:
Well, container.yml is the orchestration document. It's also the document consulted for `build` and `shipit`. I kinda figured `ansible/container.yml` made sense given the name of the project.

`main.yml` I chose based on my Ansible experience, but I'm happy to consider more useful names. You're correct that primarily used during `build`.

gregdek:
Hmm, yeah. 

"container" can mean many things. Especially when it's actually "services" -- it's a service definition file that happens to describe mostly containers.

"shipit" is weird because it uses the container.yml file, but it uses it exclusively to build other artifacts. 

Maybe it's "services.yml" and "build.yml"?

j00bar:
Wish we'd changed it a week ago if we were going to. How important is this to you?

gregdek:
Well, it got three upvotes, so people see the reasoning. It's not like we have a pypi version installable yet, right? 

I don't think it has to be done right away, but I think it does help to clarify things a bit, and we can use every advantage we can get in that regard.

darkn3rd:
How about `compose.yaml` and `build.yaml`?

mablae:
Just keep the current defaults but let the parameters be configurable on CLI? 

See twitter discussion here: https://twitter.com/mablae/status/756940374984327168

PR #207, opened 2016-08-30
implement --playbook, add tests, fix #161

### LINKED ISSUE #161: Allow to specify playbook to be used
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Currently the playbook has to be in main.yml in the ansible subfolder.
It would be greate to add teh ability to specify which playbook to use.
That would make reusing existing playbook to build images, much easier

##### EXPECTED RESULTS
Something like --playbook myplaybook.yml on the CLI.

containscafeine:
This fixes #161, allows specifying playbooks using `ansible-container --playbook <path to playbook>

### DIFF
--- a/container/cli.py
+++ b/container/cli.py
@@ -174,6 +174,10 @@ def commandline():
     parser.add_argument('--var-file', action='store',
                         help=u'Path to a YAML or JSON formatted file providing variables for '
                              u'Jinja2 templating in container.yml.', default=None)
+    parser.add_argument('--playbook', action='store',
+                           help=u'Path to playbook to use to provision '
+                                u'container images, overrides ansible/main.yml')
+
 
     subparsers = parser.add_subparsers(title='subcommand', dest='subcommand')
     for subcommand in AVAILABLE_COMMANDS:
@@ -190,6 +194,9 @@ def commandline():
     if args.debug and args.subcommand != 'version':
         LOGGING['loggers']['container']['level'] = 'DEBUG'
     config.dictConfig(LOGGING)
+
+    if args.playbook:
+        args.playbook = os.path.realpath(args.playbook)
 
     try:
         getattr(engine, u'cmdrun_{}'.format(args.subcommand))(**vars(args))
--- a/container/templates/build-docker-compose.j2.yml
+++ b/container/templates/build-docker-compose.j2.yml
@@ -2,7 +2,7 @@ ansible-container:
   # If no $DOCKER_HOST then we need to run privileged so we can access /var/run/docker.sock
   {% if not env.DOCKER_HOST %}privileged: true{% endif %}
   image: "{{ builder_img_id }}"
-  command: "/usr/local/bin/builder.sh /usr/local/bin/ansible-playbook {% if params.debug %}-vvv{% endif %} -i /etc/ansible/ansible-container-inventory.py -c docker {{ params.ansible_options | join(' ')  }} main.yml"
+  command: "/usr/local/bin/builder.sh /usr/local/bin/ansible-playbook {% if params.debug %}-vvv{% endif %} -i /etc/ansible/ansible-container-inventory.py -c docker {{ params.ansible_options | join(' ')  }} {% if params.playbook %} /tmp/playbooks/{{ params.playbook.split('/')[-1] }} {% else %} main.yml {% endif %}"
   environment:
     {% if env.DOCKER_HOST %}- DOCKER_HOST{% else %}- DOCKER_HOST=unix:///var/run/docker.sock{% endif %}
     {% if env.DOCKER_TLS_VERIFY %}- DOCKER_TLS_VERIFY{% endif %}
@@ -16,8 +16,10 @@ ansible-container:
   volumes:
     {% if not env.DOCKER_HOST %}- /var/run/docker.sock:/var/run/docker.sock{% endif %}
     {% if env.DOCKER_CERT_PATH %}- $DOCKER_CERT_PATH:/docker-certs/:Z{% endif %}
+    {% if params.playbook %}- {{ params.playbook }}:/tmp/playbooks/{{ params.playbook.split('/')[-1] }}:Z {% endif %}
     - {{ base_path }}:/ansible-container/:Z
     {% if with_volumes %}{% for vol in with_volumes %}
     - {{ vol }}{% endfor %}{% endif %}
+
   working_dir: /ansible-container/ansible/
 {{ config }}
--- a/container/engine.py
+++ b/container/engine.py
@@ -293,7 +293,7 @@ def cmdrun_build(base_path, engine_name, flatten=True, purge_last=True, local_bu
         builder_img_id = engine_obj.get_image_id_by_tag(engine_obj.builder_container_img_tag)
     except NameError:
         if local_builder:
-            create_build_container(engine_obj, base_path)
+            create_build_container(engine_obj, base_path, kwargs)
     with make_temp_dir() as temp_dir:
         logger.info('Starting %s engine to build your images...'
                     % engine_obj.orchestrator_name)
@@ -330,7 +330,7 @@ def cmdrun_build(base_path, engine_name, flatten=True, purge_last=True, local_bu
 
 
 def cmdrun_run(base_path, engine_name, service=[], production=False, **kwargs):
-    assert_initialized(base_path)
+    assert_initialized(base_path, kwargs)
     engine_args = kwargs.copy()
     engine_args.update(locals())
     engine_obj = load_engine(**engine_args)
@@ -341,7 +341,7 @@ def cmdrun_run(base_path, engine_name, service=[], production=False, **kwargs):
 
 
 def cmdrun_stop(base_path, engine_name, service=[], **kwargs):
-    assert_initialized(base_path)
+    assert_initialized(base_path, kwargs)
     engine_args = kwargs.copy()
     engine_args.update(locals())
     engine_obj = load_engine(**engine_args)
@@ -351,7 +351,7 @@ def cmdrun_stop(base_path, engine_name, service=[], **kwargs):
 
 
 def cmdrun_restart(base_path, engine_name, service=[], **kwargs):
-    assert_initialized(base_path)
+    assert_initialized(base_path, kwargs)
     engine_args = kwargs.copy()
     engine_args.update(locals())
     engine_obj = load_engine(**engine_args)
@@ -361,7 +361,7 @@ def cmdrun_restart(base_path, engine_name, service=[], **kwargs):
 
 
 def cmdrun_push(base_path, engine_name, username=None, password=None, email=None, push_to=None, **kwargs):
-    assert_initialized(base_path)
+    assert_initialized(base_path, kwargs)
     engine_args = kwargs.copy()
     engine_args.update(locals())
     engine_obj = load_engine(**engine_args)
@@ -394,7 +394,7 @@ def cmdrun_push(base_path, engine_name, username=None, password=None, email=None
 
 
 def cmdrun_shipit(base_path, engine_name, pull_from=None, **kwargs):
-    assert_initialized(base_path)
+    assert_initialized(base_path, kwargs)
     engine_args = kwargs.copy()
     engine_args.update(locals())
     engine_obj = load_engine(**engine_args)
--- a/container/templates/listhosts-docker-compose.j2.yml
+++ b/container/templates/listhosts-docker-compose.j2.yml
@@ -2,7 +2,7 @@ ansible-container:
   # If no $DOCKER_HOST then we need to run privileged so we can access /var/run/docker.sock
   {% if not env.DOCKER_HOST %}privileged: true{% endif %}
   image: "{{ builder_img_id }}"
-  command: /usr/local/bin/builder.sh /usr/local/bin/ansible-playbook -i /etc/ansible/ansible-container-inventory.py -c docker {{ params.ansible_options | join(' ')  }} --list-hosts main.yml
+  command: /usr/local/bin/builder.sh /usr/local/bin/ansible-playbook -i /etc/ansible/ansible-container-inventory.py -c docker {{ params.ansible_options | join(' ')  }} --list-hosts {% if params.playbook %} /tmp/playbooks/{{ params.playbook.split('/')[-1] }} {% else %} main.yml {% endif %}
   environment:
     {% if env.DOCKER_HOST %}- DOCKER_HOST{% else %}- DOCKER_HOST=unix:///var/run/docker.sock{% endif %}
     {% if env.DOCKER_TLS_VERIFY %}- DOCKER_TLS_VERIFY{% endif %}
@@ -13,7 +13,9 @@ ansible-container:
   volumes:
     {% if not env.DOCKER_HOST %}- /var/run/docker.sock:/var/run/docker.sock{% endif %}
     {% if env.DOCKER_CERT_PATH %}- $DOCKER_CERT_PATH:/docker-certs/:Z{% endif %}
+    {% if params.playbook %}- {{ params.playbook }}:/tmp/playbooks/{{ params.playbook.split('/')[-1] }}:Z {% endif %}
     - {{ base_path }}:/ansible-container/:Z
+
   working_dir: /ansible-container/ansible/
   stdin_open: true
   tty: true
--- a/container/utils.py
+++ b/container/utils.py
@@ -80,11 +80,12 @@ def config_format_version(base_path, config_data=None):
         config_data = get_config(base_path)
     return int(config_data.pop('version', 1))
 
-def assert_initialized(base_path):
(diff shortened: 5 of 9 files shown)


chouseknecht:
This relates to issue #76.

container/templates/build-docker-compose.j2.yml
```diff
@@ -18,5 +18,6 @@ ansible-container:
     - {{ base_path }}:/ansible-container/:Z
     {% if with_volumes %}{% for vol in with_volumes %}
     - {{ vol }}{% endfor %}{% endif %}
+    {% if params.playbook %}- {{ params.playbook }}:/ansible-container{{ params.playbook }}:Z {% endif %}
```
chouseknecht: The example from #76 is `ansible-container build --playbook="/root/ansible/playbook.yml"`. I'm curious if you have tested it? I know you're still working on this, and I'm jumping ahead.

You might also test with "/etc/ansible/playbook.yml", as "/etc/ansible" is a default path used by Ansible for storing content. The point is to test with something owned by 'root' and not the user,  and observe how permissions behave. 

Also, speaking of permissions, @dustymabe needs to review the use of `:Z` here. It's my understanding we may not want to continue using `:Z`. I'm not an SELinux expert, so I may be wrong. He should at least be aware of it though.

dustymabe: > Also, speaking of permissions, @dustymabe needs to review the use of :Z here. It's my understanding we may not want to continue using :Z. I'm not an SELinux expert, so I may be wrong. He should at least be aware of it though.

the use of `:Z` should be fine in this case as it is consistent with what is done elsewhere right now in ansible-container. As part of https://github.com/ansible/ansible-container/issues/168 we may overhaul all of the uses of `:Z` and replace with some conditional logic to determine if `:Z` should be used or not.

containscafeine: @chouseknecht yep, checked against playbooks owned by root, it works.

@dustymabe gotcha, `:Z` remains for now.

chouseknecht:
Testing this now. Apologies for the delay. Will either have comments or merge it today.

chouseknecht:
@containscafeine 

Having an issue using the playbook from any location. It also breaks `ansible-container --debug version`. I'm running Docker Machine 1.11.2 on Mac.

Here's output from `ansible-container --debug version`:

```
$ ac --debug version
Ansible Container, version 0.2.0-pre
Darwin, Chriss-MacBook-Pro.local, 15.6.0, Darwin Kernel Version 15.6.0: Thu Jun 23 18:25:34 PDT 2016; root:xnu-3248.60.10~1/RELEASE_X86_64, x86_64
2.7.11 (default, Jan 22 2016, 08:29:18)
[GCC 4.2.1 Compatible Apple LLVM 7.0.2 (clang-700.1.81)] /usr/local/opt/python/bin/python2.7
No Ansible Container project data found - do you need to run "ansible-container init"?
```


Here's the debug output from the`build`:
```
$ ac --debug --playbook /etc/ansible/build.yml build
Project name is custom_playbook
Default vars:
{}
Getting environment variables...
Template variables:
 {}
Config:
{
    "registries": {},
    "services": {
        "web": {
            "image": "alpine"
        }
    },
    "version": "1"
}
Initialized with params: {'with_volumes': None, 'local_builder': False, 'with_variables': None, 'rebuild': False, 'subcommand': 'build', 'var_file': None, 'engine_args': {'local_builder': False, 'subcommand': 'build', 'base_path': '/Users/chouseknecht/projects/custom_playbook', 'ansible_options': u'', 'playbook': '/private/etc/ansible/build.yml', 'no_purge_last': True, 'with_volumes': None, 'engine_name': 'docker', 'with_variables': None, 'rebuild': False, 'var_file': None, 'engine_args': {...}, 'save_build_container': False, 'kwargs': {'with_volumes': None, 'with_variables': None, 'subcommand': 'build', 'var_file': None, 'playbook': '/private/etc/ansible/build.yml', 'debug': True, 'no_purge_last': True}, 'debug': True, 'purge_last': True, 'flatten': False}, 'save_build_container': False, 'purge_last': True, 'ansible_options': u'', 'playbook': '/private/etc/ansible/build.yml', 'kwargs': {'with_volumes': None, 'with_variables': None, 'subcommand': 'build', 'var_file': None, 'playbook': '/private/etc/ansible/build.yml', 'debug': True, 'no_purge_last': True}, 'debug': True, 'no_purge_last': True, 'flatten': False}
Using temporary directory /var/folders/g_/mr68c1yn1rl4b7qh6pntbmmm0000gn/T/tmpB3cDFg...
Starting Docker Compose engine to build your images...
Using temporary directory /var/folders/g_/mr68c1yn1rl4b7qh6pntbmmm0000gn/T/tmpziMfB3...
Compose derived from config:
{'web': {'image': 'alpine'}}
{'web': {'command': 'sh -c "while true; do sleep 1; done"', 'image': 'alpine', 'entrypoint': [], 'working_dir': '/', 'user': 'root'}}
Config YAML is
web:
  command: sh -c "while true; do sleep 1; done"
  entrypoint: []
  image: alpine
  user: root
  working_dir: /

Rendered Jinja Template:
ansible-container:
  # If no $DOCKER_HOST then we need to run privileged so we can access /var/run/docker.sock

  image: "ansible/ansible-container-builder:0.2"
  command: /usr/local/bin/builder.sh /usr/local/bin/ansible-playbook -i /etc/ansible/ansible-container-inventory.py -c docker  --list-hosts  /tmp/playbooks/build.yml
  environment:
    - DOCKER_HOST
    - DOCKER_TLS_VERIFY
    - DOCKER_CERT_PATH=/docker-certs/
    - COMPOSE_HTTP_TIMEOUT=3000
    - DOCKER_API_VERSION=1.23
    - ANSIBLE_ORCHESTRATED_HOSTS=web
  volumes:

    - $DOCKER_CERT_PATH:/docker-certs/:Z
    - /private/etc/ansible/build.yml:/tmp/playbooks/build.yml:Z
    - /Users/chouseknecht/projects/custom_playbook:/ansible-container/:Z

  working_dir: /ansible-container/ansible/
  stdin_open: true
  tty: true
web:
  command: sh -c "while true; do sleep 1; done"
  entrypoint: []
  image: alpine
  user: root
  working_dir: /

Cleaning up Ansible Container builder...
--list-hosts
ansible-container_1  | ERROR! the playbook: /tmp/playbooks/build.yml does not appear to be a file
ansible_ansible-container_1 exited with code 1

Cleaning up temporary directory /var/folders/g_/mr68c1yn1rl4b7qh6pntbmmm0000gn/T/tmpziMfB3...
Compose derived from config:
{'web': {'image': 'alpine'}}
Using temporary directory /var/folders/g_/mr68c1yn1rl4b7qh6pntbmmm0000gn/T/tmpaWuUSD...
Compose derived from config:
{'web': {'image': 'alpine'}}
{'web': {'command': 'sh -c "while true; do sleep 1; done"', 'image': 'alpine', 'entrypoint': [], 'working_dir': '/', 'user': 'root'}}
Config YAML is
web:
  command: sh -c "while true; do sleep 1; done"
  entrypoint: []
  image: alpine
  user: root
  working_dir: /

Rendered Jinja Template:
ansible-container:
  # If no $DOCKER_HOST then we need to run privileged so we can access /var/run/docker.sock

  image: "ansible/ansible-container-builder:0.2"
  command: /usr/local/bin/builder.sh /usr/local/bin/ansible-playbook -i /etc/ansible/ansible-container-inventory.py -c docker  --list-hosts  /tmp/playbooks/build.yml
  environment:
    - DOCKER_HOST
    - DOCKER_TLS_VERIFY
    - DOCKER_CERT_PATH=/docker-certs/
    - COMPOSE_HTTP_TIMEOUT=3000
    - DOCKER_API_VERSION=1.23
    - ANSIBLE_ORCHESTRATED_HOSTS=web
  volumes:

    - $DOCKER_CERT_PATH:/docker-certs/:Z
    - /private/etc/ansible/build.yml:/tmp/playbooks/build.yml:Z
    - /Users/chouseknecht/projects/custom_playbook:/ansible-container/:Z

  working_dir: /ansible-container/ansible/
  stdin_open: true
  tty: true
web:
  command: sh -c "while true; do sleep 1; done"
  entrypoint: []
  image: alpine
  user: root
  working_dir: /

Cleaning up Ansible Container builder...
--list-hosts
ansible-container_1  | ERROR! the playbook: /tmp/playbooks/build.yml does not appear to be a file
ansible_ansible-container_1 exited with code 1

Cleaning up temporary directory /var/folders/g_/mr68c1yn1rl4b7qh6pntbmmm0000gn/T/tmpaWuUSD...
Orchestrated hosts: []
No NameError raised when searching for tag custom_playbook-web:latest
{'web': {'image': 'custom_playbook-web:latest'}}
Config YAML is
web: {image: 'custom_playbook-web:latest'}

Rendered Jinja Template:
ansible-container:
  # If no $DOCKER_HOST then we need to run privileged so we can access /var/run/docker.sock

  image: "ansible/ansible-container-builder:0.2"
  command: "/usr/local/bin/builder.sh /usr/local/bin/ansible-playbook -vvv -i /etc/ansible/ansible-container-inventory.py -c docker   /tmp/playbooks/build.yml "
  environment:
    - DOCKER_HOST
    - DOCKER_TLS_VERIFY
    - DOCKER_CERT_PATH=/docker-certs/
    - COMPOSE_HTTP_TIMEOUT=3000
    - DOCKER_API_VERSION=1.23
    - ANSIBLE_ORCHESTRATED_HOSTS=web

  volumes:

    - $DOCKER_CERT_PATH:/docker-certs/:Z
    - /private/etc/ansible/build.yml:/tmp/playbooks/build.yml:Z
    - /Users/chouseknecht/projects/custom_playbook:/ansible-container/:Z


  working_dir: /ansible-container/ansible/
web: {image: 'custom_playbook-web:latest'}

Attaching to ansible_web_1, ansible_ansible-container_1
web_1                | /bin/sh: -c requires an argument
ansible_web_1 exited with code 2
Aborting on container exit...
Stopping ansible_ansible-container_1 ... done
Ansible playbook run failed.
Cleaning up Ansible Container builder...
Cleaning up temporary directory /var/folders/g_/mr68c1yn1rl4b7qh6pntbmmm0000gn/T/tmpB3cDFg...
Ansible build failed
  File "/Users/chouseknecht/projects/ansible-container/container/cli.py", line 202, in commandline
    getattr(engine, u'cmdrun_{}'.format(args.subcommand))(**vars(args))
  File "/Users/chouseknecht/projects/ansible-container/container/engine.py", line 320, in cmdrun_build
    raise RuntimeError(u'Ansible build failed')
RuntimeError: Ansible build failed
```

chouseknecht:
@containscafeine 
It seems that Docker Machine first mounts the host portion of the `volume` directive to the vagrant box, and then mounts that to the container. 

So Ansible Container requested `volume: /etc/ansible/build.yml:/tmp/playbooks/build.yml`, but Docker Machine did not recognize /etc/ansible/build.yml as an absolute path to a file. Instead, it treated it as a directory when mounting it to the vagrant box.

Here what's on my vagrant box:

```
docker@default:/private/etc/ansible$ ls -l
total 0
drwxr-xr-x    2 root     root            40 Sep  6 20:08 build.yml/
```

I think we're thinking about this incorrectly.  The goal is to enable Ansible users to use their existing content (playbooks, roles, etc.), not just a single playbook. I think we started down the right path adding the`--with-volumes` option. In theory I should be able to mount all my directory(s) of playbooks and roles, and then call a particular playbook from one of those volumes from within the build container. 

Maybe `--playbook` should not be attempting to mount a volume. It should assume you will provide the appropriate volume using `---with-volume`, and the path you provide should be relative to a provided volume.

chouseknecht:
@containscafeine If I don't hear back from you in a few hours, I'll probably close this PR and work on a new version. We also need to build up a good example that uses existing playbooks and roles.

I'm also thinking about adding a new top-level section to container.yml that provides a way to permanently set volumes, environment vars, playbook name for the build container. The list operation could grab it as well. This would be a convenience, making it so that the same command line options do not have to be continually repeated.

containscafeine:
@chouseknecht Thanks for going through this, I got to go through your comments only now.

> Having an issue using the playbook from any location. It also breaks ansible-container --debug version

Does running `version` need to check if the current directory has been initialized anyway?
The same error occurs when on the develop branch if we are not in an initialized directory.

> `ansible-container_1  | ERROR! the playbook: /tmp/playbooks/build.yml does not appear to be a file`
> ...
> It seems that Docker Machine first mounts the host portion of the volume directive to the vagrant box, and then mounts that to the container.
> So Ansible Container requested volume: /etc/ansible/build.yml:/tmp/playbooks/build.yml, but Docker Machine did not recognize /etc/ansible/build.yml as an absolute path to a file. Instead, it treated it as a directory when mounting it to the vagrant box.

This is weird, it works fine here on my Fedora box and tests pass too. As you mentioned, this seems to be a Mac specific error and I don't have one to test this on. Any pointers how this should be done?

> I think we're thinking about this incorrectly. The goal is to enable Ansible users to use their existing content (playbooks, roles, etc.), not just a single playbook.

I am looking at this PR to allow only one playbook (as a replacement to having a mandatory `ansible/main.yml`). As for roles, when the playbook is parsed, the entire role's hierarchy should get mounted inside the builder container and used to provision the container. How about adding that logic during the build process?

>  I think we started down the right path adding the --with-volumes option. In theory I should be able to mount all my directory(s) of playbooks and roles, and then call a particular playbook from one of those volumes from within the build container.
> Maybe --playbook should not be attempting to mount a volume. It should assume you will provide the appropriate volume using ---with-volume, and the path you provide should be relative to a provided volume.

Well, in this approach the user has to be aware that there is something like a builder container and the volumes are mounted inside that and then those are used to provision my container image. This does not seem to be a direct replacement for `main.yml`.

From an end user perspective, it's just easier to specify `--playbook /etc/ansible/playbook.yml` than having to pass multiple parameters.

> I'm also thinking about adding a new top-level section to container.yml that provides a way to permanently set volumes, environment vars, playbook name for the build container. The list operation could grab it as well. This would be a convenience, making it so that the same command line options do not have to be continually repeated.

I see `container.yml` being more of an application definition and specific to containers being built and services being created. About the builder container specific parameters, how about picking those up from a local .cfg file? But yep, makes sense to not having command line options repeated over and over.

Thoughts? Do I make sense?

chouseknecht:
@containscafeine 

>I am looking at this PR to allow only one playbook (as a replacement to having a mandatory ansible/main.yml). As for roles, when the playbook is parsed, the entire role's hierarchy should get mounted inside the builder container and used to provision the container. How about adding that logic during the build process?

The problem is definitely larger than mounting just one playbook. And now that I'm reading your comments, I think what you're proposing makes sense. 

If --playbook returns /etc/ansible/foo.yml, then we should:

- mount /etc/ansible:/etc/ansible 
- override main.yml with /etc/ansible/foo.yml

If --playbook returns <some path already mounted to the container>/foo.yml, then we should:

- detect and not duplicate the mount
- override main.yml with <some path already mounted to the container>/foo.yml

### Additional requirements...

- The build container already relies on the inventory script being in /etc/ansible, so we need to somehow make sure the inventory script does not get lost should the user create an /etc/ansible volume. This probably means we need to put the inventory script somewhere else within the build container. 
-  Mounting things into Docker Machine outside of the user's home directly seems to be nearly impossible. This is something we need to thoroughly document as lots of people are running Docker Machine for development.

### PR #216
I started PR #216 to address these issues, plus we need to move --with-volumes and --with-vars. to be top-level parameters. They need to be available for both `build` and `run`, otherwise the operation to determine which hosts are touched by the playbook fails.

### `container.yml` or a config file... 

I really don't want to introduce a third file. Maybe what makes sense is to allow the user to add a service named `ansible-container`, and that's where we accept overrides to specific things like playbook (which would be a new directive), environment and volumes. And then at build time, we merge the user's `ansible-container` settings into the actual `ansible-container` service.

container/cli.py
```diff
@@ -174,6 +174,10 @@ def commandline():
     parser.add_argument('--var-file', action='store',
                         help=u'Path to a YAML or JSON formatted file providing variables for '
                              u'Jinja2 templating in container.yml.', default=None)
+    parser.add_argument('--playbook', action='store',
```
j00bar: Maybe use `default` and not have to list it in the help text?

container/cli.py
```diff
@@ -191,6 +195,9 @@ def commandline():
         LOGGING['loggers']['container']['level'] = 'DEBUG'
     config.dictConfig(LOGGING)
 
+    if args.playbook:
+        args.playbook = os.path.realpath(args.playbook)
```
j00bar: Not sure you want to do this, as it needs to be evaluated in the context of the build container.

j00bar:
Sorry - started reviewing and then realized that this is effectively superseded by #216

chouseknecht:
Closing. We've decided that re-using existing Ansible content should be focused exclusively on Ansible roles, and we mostly already have a solution to that.

PR #216, opened 2016-09-07
Make existing Ansible content usable by adding --playbook option and …

chouseknecht:
##### ISSUE TYPE
Feature Pull Request

##### SUMMARY
**WIP - do NOT merge**

The following changes create a demonstrable path for re-using existing Ansible content:
 
- Add --playbook option for providing the path to a playbook that exists within the Ansible build container.
- Does not mount the --playbook path, but instead relies on the user to mount the appropriate path using --with-volumes. Based on user feedback, users typically have an existing directory of playbooks that need to be accessible from within the build container.
- Changes --with-volumes and --with-variables to be top level options allowing them to be accessible during both the `build` command and  the `run` command.

TODO:

- Once the above is working, then add a top level section in `container.yml` for specifying the build options.
- Fix existing tests
- Add new tests
- Update docs
- Add a `Using existing playbooks and roles` page or topic on the docs site that explains how it works, and limitations/known issues when using Docker Machine.
 
Here's basically how you can mount an existing playbook dir and role dir, and use a playbook from the playbook dir:

ansible-container --playbook /playbooks/httpd.yml \
--with-volumes ~/projects/playbooks:/playbooks \ 
--with-volumes ~/projects/roles:/roles \
build

This approach gives the user complete control over path names as well. They can also place an ansible.cfg file in the project `ansible` directory to control the `roles_path` and any other settings.

### DIFF
--- a/container/docker/engine.py
+++ b/container/docker/engine.py
@@ -85,7 +85,7 @@ class Engine(BaseEngine):
 
         :return: generator of strings
         """
-        assert_initialized(self.base_path)
+        assert_initialized(self.base_path, playbook=self.playbook)
         client = self.get_client()
         with make_temp_dir() as temp_dir:
             logger.info('Building Docker Engine context...')
@@ -757,6 +757,13 @@ class Engine(BaseEngine):
         config_yaml = yaml_dump(config)
         logger.debug('Config YAML is')
         logger.debug(config_yaml)
+        with_volumes = self.params.get('with_volumes')
+        with_variables = self.params.get('with_variables')
+
+        logger.debug("with_volumes: %s" % with_volumes)
+
+        #logger.debug("with volumes: {0}\n with_variables: {1}".format(','.join(with_volumes),
+        #                                                              ','.join(with_variables)))
         jinja_render_to_temp('%s-docker-compose.j2.yml' % (operation,),
                              temp_dir,
                              'docker-compose.yml',
@@ -768,6 +775,9 @@ class Engine(BaseEngine):
                              builder_img_id=builder_img_id,
                              config=config_yaml,
                              env=os.environ,
+                             playbook=self.playbook,
+                             with_volumes=with_volumes,
+                             with_variables=with_variables,
                              **context)
         options = self.DEFAULT_COMPOSE_OPTIONS.copy()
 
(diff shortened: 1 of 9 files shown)


container/docker/engine.py
```diff
@@ -768,6 +775,9 @@ def bootstrap_env(self, temp_dir, behavior, operation, compose_option,
                              builder_img_id=builder_img_id,
                              config=config_yaml,
                              env=os.environ,
+                             playbook=self.playbook,
+                             with_volumes=with_volumes,
```
j00bar: Since `with_volumes` and `with_variables` are in `self.params`, which is also passed as an argument, no need to repeat them here.

container/utils.py
```diff
@@ -57,23 +58,47 @@ def jinja_render_to_temp(template_file, temp_dir, dest_file, **context):
...
 def config_format_version(base_path, config_data=None):
     if not config_data:
         config_data = get_config(base_path)
     return int(config_data.pop('version', 1))
 
-def assert_initialized(base_path):
-    ansible_dir = os.path.normpath(
-        os.path.join(base_path, 'ansible'))
+
+def assert_initialized(base_path, kwargs):
```
j00bar: Using `kwargs` here makes for a messy method signature - since it's `playbook` that you're interested in as a kwarg, just use `playbook='main.yml'` here.

container/docker/engine.py
```diff
@@ -85,7 +85,7 @@ def build_buildcontainer_image(self):
 
         :return: generator of strings
         """
-        assert_initialized(self.base_path)
+        assert_initialized(self.base_path, playbook=self.playbook)
```
j00bar: `playbook` is not a valid parameter to `assert_initialized`

j00bar: Of course, if you follow my suggestion to change the method signature of `assert_initialized`, this can be left alone.

j00bar:
Good approach, if you're going to do this. I still think this is not the best idea in general. It should be a Good Thing™ to keep your playbook in the same VCS repository as your code, not elsewhere.

chouseknecht:
If starting a project from scratch, I agree, just use main.yml. But, we keep getting asked, "How do I use my existing playbooks and roles."

container/docker/engine.py
chouseknecht: Yes, it is in params. Breaking it out of params and passing in an explicit value makes the template less icky.

chouseknecht:
Closing. We've decided that re-using existing Ansible content should be focused exclusively on Ansible roles, and we mostly already have a solution to that.

PR #238, opened 2016-09-19
edit requirements.txt.

procrypt:
##### ISSUE TYPE
- Bug Pull Request
##### SUMMARY

I was trying build the RPM for ansible-containers and I figured out that it makes it difficult to install the RPM, if we have such rigid coupling of dependencies.
If we update the requirements.txt to not look for the exact versions of dependencies but something like,

```
docker-compose>=1.7.0
```

 then we can RPM install successfully.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,3 @@
-docker-compose==1.7.0
-docker-py==1.8.0
-Jinja2==2.8
-PyYAML==3.11
-requests==2.7.0
+docker-compose>=1.7.0
+Jinja2>=2.8
+PyYAML>=3.11


procrypt:
CC @dustymabe.

dustymabe:
LGTM

@chouseknecht PTAL - basically this makes it easier for us to pull the source code and not have to modify it before creating an rpm out of it. There may be reasons you don't want to make this change upstream though.

dustymabe:
hey @j00bar would it be reasonable to remove the deps that are pulled in by docker-compose in our requirements.txt? i.e. we would trim it down to something like the following since docker-compose pulls in docker-py and requests already: https://github.com/docker/compose/blob/master/requirements.txt

```
$ cat requirements.txt 
docker-compose>=1.7.0
Jinja2>=2.8
PyYAML>=3.11
```

dustymabe:
@j00bar PTAL at the last comment

chouseknecht:
The proposed change to `requirements.txt` is fine. We can merge this once tests pass.

dustymabe:
@procrypt please update the requirements and test and then push up the changes so it can run through tests here.

dustymabe:
> @procrypt please update the requirements and test and then push up the changes so it can run through tests here.

err - nevermind. looks like you already did that. I'm looking at the test failure now.

dustymabe:
see https://github.com/ansible/ansible-container/pull/245#issuecomment-249069367

dustymabe:
I opened https://github.com/ansible/ansible-container/pull/254 to hopefully take care of this if the tests pass this time.

dustymabe:
since we aren't going to do this now let's close this: see #259

PR #219, opened 2016-09-08
[WIP] moving to compose v2

containscafeine:
- This is a very, very rough draft of the PR.
- Please do not review/merge.
## 
- [x] Check if Docker Compose v2 files work
- [x] Check if all the other subcommands work fine
- [ ] Check if `shipit` works file with this
- [ ] Add mappings for additional directives to `shipit`
- [x] Add tests for exhaustive Compose v2 files
- [x] Add tests for Compose v1 file (backward compatibility) - [already there]
- [ ] Don't break stuff
- [ ] Document which compose directives work, how they work, and which ones don't work

PR #239, opened 2016-09-20
shipit --save config via kompose, fix #233

### LINKED ISSUE #233: Make shipit do Kompose things
It occurred to me that the issue with using Kompose today is a lack of package management. If we could easily and reliably install it, we could just call it from Python as a subprocess. 

Turns out we have a package manager, so we could have a versioned role in Galaxy where the role version corresponds to the Kompose version. Also, we can bake it into the build container image. If you run with --local build, we install it via a role, otherwise what's in the image gets used. And, if you want to override the version being auto installed or shipped in the image, just add the role and version to your project's requirements.yml
 
Now we just need to agree on how we want to use it, and what to do with the current 'shipit' state. I'll put some thought into that soon, but in the meantime comments and suggestions welcome.

containscafeine:
##### ISSUE TYPE
- Feature Pull Request
##### SUMMARY
- Added another layer on top of the builder container by adding the kompose binary, which is now used to generate Kubernetes and OpenShift artifacts when `ansible-container shipit x --save-config` is run.
- First the directory in which the parsed container.yml is stored is mounted inside the builder container, and then `kompose` does the conversion and spits out the artifacts.

TODO:
- [x] Add check if the builder image with the kompose tag is present or not, if not then pull it.
- [x] Add tests
## 

Unblocks #219, Fixes #233

### DIFF
--- a/container/shipit/kubernetes/engine.py
+++ b/container/shipit/kubernetes/engine.py
@@ -2,21 +2,28 @@
 
 from __future__ import absolute_import
 
+from compose.cli.command import project_from_options
+from compose.cli import main
 import logging
 import os.path
-import json
+import yaml
 
 from ..base_engine import BaseShipItEngine
 from .deployment import Deployment
 from .service import Service
 from ..utils import create_path
 from ..constants import SHIPIT_PATH, SHIPIT_CONFIG_PATH
+from ...utils import jinja_render_to_temp
+from ...temp import MakeTempDir as make_temp_dir
+from ...docker.engine import Engine
 
 logger = logging.getLogger(__name__)
 
 
 class ShipItEngine(BaseShipItEngine):
     name = u'kubernetes'
+    builder_container_img = 'ansible-container-builder'
+    builder_container_img_tag = '0.2-kompose-v0.1.0'
 
     def add_options(self, subparser):
         super(ShipItEngine, self).add_options(subparser)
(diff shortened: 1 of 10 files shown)


container/templates/ansible-kompose-dockerfile.j2
```diff
@@ -0,0 +1,7 @@
+FROM ansible/ansible-container-builder:0.2
+
+# Fetch kompose
+ADD https://github.com/skippbox/kompose/releases/download/v0.1.0/kompose_linux-amd64.tar.gz /tmp/
+
+# Decompress and add binary to $PATH
+RUN tar xzf /tmp/kompose_linux-amd64.tar.gz --strip-components 1 -C /usr/bin/
```
dustymabe: would probably be better if you run these in a single step and then delete /tmp/kompose_linux-amd64.tar.gz so that file doesn't get committed into a layer

containscafeine: Fixed

container/templates/ansible-kompose-dockerfile.j2
```diff
@@ -0,0 +1,4 @@
+FROM ansible/ansible-container-builder:0.2
+
+# Fetch kompose, decomporess, add binary to $PATH
```
dustymabe: s/decomporess/decompress/

container/templates/ansible-kompose-dockerfile.j2
```diff
@@ -0,0 +1,4 @@
...
+RUN curl -L https://github.com/skippbox/kompose/releases/download/v0.1.0/kompose_linux-amd64.tar.gz -o /tmp/kompose_linux-amd64.tar.gz && tar xzf /tmp/kompose_linux-amd64.tar.gz --strip-components 1 -C /usr/bin/ && rm -f  /tmp/kompose_linux-amd64.tar.gz
```
dustymabe: can we break this line up with some `\` on the end?

test/local/ansible/container.yml
```diff
@@ -5,6 +5,7 @@ services:
     volumes:
        - "${ANSIBLE_CONTAINER_PATH}:${ANSIBLE_CONTAINER_PATH}" 
        - /var/run/docker.sock:/var/run/docker.sock
+       - /tmp/:/tmp/
```
dustymabe: what was the point of this one?

dustymabe:
was trying to test this out and i'm seeing an odd error. seen this before? 

```
$ ansible-container shipit  kube --save-config 
No handlers could be found for logger "container.docker.engine"
Unable to determine namespace for registry https://index.docker.io/v1/. Error: Unable to read your docker config file. Try providing login credentials for the registry.. Either authenticate with the registry or provide a namespace for the registry in container.yml
```

I was trying to test against https://github.com/ansible/ansible-container-examples/tree/master/wordpress

container/templates/ansible-kompose-dockerfile.j2
containscafeine: Fixed

container/templates/ansible-kompose-dockerfile.j2
containscafeine: Fixed

containscafeine:
@dustymabe -

> $ ansible-container shipit  kube --save-config 
No handlers could be found for logger "container.docker.engine"
Unable to determine namespace for registry https://index.docker.io/v1/. Error: Unable to read your docker config file. Try providing login credentials for the registry.. Either authenticate with the registry or provide a namespace for the registry in container.yml

IIRC, you need to be authenticated to a registry to use `shipit`, so if you do a `docker login`, it should work.

container/shipit/kubernetes/engine.py
```diff
@@ -33,19 +40,41 @@ def save_config(self):
...
 
-        templates = Deployment(config=self.config, project_name=self.project_name).get_template()
-        for template in templates:
-            name = "%s-deployment.json" % template['metadata']['name']
-            with open(os.path.join(dest_path, name), 'w') as f:
-                f.write(json.dumps(template, indent=4))
+            jinja_render_to_temp('shipit-kompose-docker-compose.j2.yml',
+                                 temp_dir,
+                                 'shipit-kompose-compose.yml',
+                                 builder_img_id="containscafeine/{}:{}".format(
```
dustymabe: we're going to need to fix this to be not from your account. either that or we actually just install kompose into the ansible-container builder image - it will be in epel soon so we can just install from rpm. 

@chouseknecht, thoughts?

containscafeine: Yep, once @chouseknecht gives this a green flag and pushes the build image, I will point this to ansible's docker hub.

PR #241, opened 2016-09-21
builder: move container builder to a centos base image

dustymabe:


dustymabe:
@chouseknecht PTAL - if we merge this change we'll need to build/push a new version of the ansible-container image to docker hub

dustymabe:
@chouseknecht - it also seems to me like the ansible-container builder is just a generic image that has ansible installed - should we somehow rename it to something more generic and open it up for wider uses as a "ansible engine" for people to use?

dustymabe:
ok so the tests failed because the tests pull the ansible-container-builder from docker hub - I'm going to push an extra commit that tells it to do a local build. Not sure if this is what we want to do long term or not, but we need to account for changes in the container builder image somehow.

chouseknecht:
@dustymabe lets leave it as `ansible-container-builder` for now. The point of the image is to support Ansible Container, and it's tailored specifically for that.  If it happens to work for other things, then bonus.

Will we ever have other `builders`? Maybe. I don't know. Let's cross that bridge when we get there.

PR #242, opened 2016-09-22
A bunch of fixes in final image metadata

gnosek:
##### ISSUE TYPE
 - Bugfix Pull Request

##### SUMMARY
The issues fixed by this pull request (as a single PR, as they all touch the same code) are:

 - images may not have a custom entrypoint specified (it is ignored)
 - images may not have a command specified as a list (only as a string that gets passed to the shell)
 - images have labels left over from `docker-compose` which means containers created from them get destroyed at every `ansible-composer` run

### DIFF
--- a/container/docker/engine.py
+++ b/container/docker/engine.py
@@ -524,9 +524,13 @@ class Engine(BaseEngine):
         cmd = self.config['services'][host].get('command', '')
         if isinstance(cmd, list):
             cmd = json.dumps(cmd)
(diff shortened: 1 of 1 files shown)


chouseknecht:
@gnosek 

Hey, thank you for your contribution! Really appreciate you taking the time to dig into this and help us out. We're moving as fast as we can, but we can't do it alone. 

Thank you !

gnosek:
Glad to help, I was just scratching my itch :)

PR #243, opened 2016-09-22
builder: add symlink to ease transition

dustymabe:
In 9deb3eb we moved to /usr/bin/ansible-playbook
Let's ease the transition on people using <9deb3eb code
by symlinking to the new place


dustymabe:
damn. now how did that break tests. /me looking

dustymabe:
weird. initially the tests showed as failed and then they showed as passed. *suspicious emoji*

PR #245, opened 2016-09-22
this is a test - do not merge

dustymabe:


dustymabe:
So this PR is exactly like #238 except with a modification to print out more info (stdout/stderr) in the failed test. stdout has the [good contents](https://travis-ci.org/ansible/ansible-container/builds/162050816#L751) (if we were checking stdout the test would pass). stderr looks to have [corrupted content](https://travis-ci.org/ansible/ansible-container/builds/162050816#L850)

any ideas?

chouseknecht:
I made some changes to the ansible-docker.j2 template. It was installing docker-engine 1.10.3, which failed to respect DOCKER_CERT_PATH. It now install 1.11.2, and tests seem to be working both locally and on Travis.

test/integration/test_slow.py
```diff
@@ -127,7 +127,10 @@ def test_install_role_requirements():
     env = ScriptTestEnvironment()
     result = env.run('ansible-container', '--debug', 'build',
                      cwd=project_dir('requirements'), expect_stderr=True)
-    assert "ansible-role-apache was installed successfully" in result.stderr
+    print result.stdout
+    print "YYYYYYYYYYYYYYYYYYYYYYYYYYY\n"
```
chouseknecht: We're going from a known working versions of docker-compose, docker-py, Jinja2 and PyYaml to just installing the latest. It seems that installing the latest of one or more of these packages broke something, because this is the only place where this test fails.

Can you please dig a little deeper and figure out what's actually going on. I don't think inserting a hack just to make the test work should be an acceptable solution. It seems to me that we run the risk of having this bug surface in places where the hack is not implemented.

requirements.txt
```diff
@@ -1,5 +1,3 @@
-docker-compose==1.7.0
-docker-py==1.8.0
-Jinja2==2.8
-PyYAML==3.11
-requests==2.7.0
+docker-compose>=1.7.0
+Jinja2>=2.8
+PyYAML>=3.11
```
chouseknecht: Remind me again why it's not OK to pin these packages to versions that are known to work?

PR #249, opened 2016-09-23
Roles and Galaxy integration

j00bar:
* Remove example from codebase, in favor of Galaxy-hosted container app
* Galaxy integration through `install` and `init` commands.
* `--roles-path` command-line argument
* Simplified project template
* All the docs


chouseknecht:
@j00bar - this needs rebasing.

PR #254, opened 2016-09-26
edit requirements

dustymabe:


dustymabe:
I'm not sure why I had to update the test but it was some weird behavior with the checking of the output streams in the test. I documented/commented on that behavior in this comment: https://github.com/ansible/ansible-container/pull/245#issuecomment-249069367

requirements.txt
```diff
@@ -1,5 +1,3 @@
-docker-compose==1.7.0
-docker-py==1.8.0
-Jinja2==2.8
-PyYAML==3.11
-requests==2.7.0
+docker-compose>=1.7.0
+Jinja2>=2.8
+PyYAML>=3.11
```
chouseknecht: Remind me why we are moving from known working versions of these modules to installing the latest? It's not uncommon for a breaking change to be released in Docker at which point we may be forced to pin the version docker-compose.

dustymabe: when creating an rpm we can't tie to specific versions because we can't exactly dictate what versions of each dependency will exist in the OS distribution. This means when creating an RPM we'll have to have a requirements.txt that looks something like this.

Now we do have options here though. One option was to try to get upstream (here on github) to implement this change upstream in the code. The other option is to patch the requirements.txt file during RPM creation. We were attempting to go with the first option, but it's not required as we can patch this in the RPM if you would prefer that.

chouseknecht: And... why are we building an RPM?

dustymabe: Because it is the way we distribute software for Fedora/CentOS/RHEL. Most people (myself included) would not like to install things via pip or other language based package managers.

Do you not see value in building an RPM ?

gregdek: Hard to argue what "most people" would do nowadays. :)

That said: we shouldn't preclude our friends in Fedora-land from carrying a patched RPM if they see fit -- but we should be *very* clear, as upstream, what versions of these things we support. And for now, we want to pin these versions.

dustymabe: Talked with @gregdek. Right now the best argument against having an RPM is that Ansible-Container is not stable enough to warrant having an RPM. We'll post up the work we have done to create an Ansible Container RPM so that work can be picked up when the time is right.

dustymabe: @chouseknecht is there a place on the roadmap we could put this with a link to the work we have done? Where does that roadmap live?

chouseknecht: @dustymab, open an issue to track the need for an RPM and the work done to date. When the time is right, we'll add it to a roadmap. Roadmap docs live [here](https://github.com/ansible/ansible-container/tree/develop/docs/rst/roadmaps).

dustymabe: I opened #254 for this.

dustymabe:
closing until the time is right for an rpm - see #259

PR #257, opened 2016-09-27
Remove unused var and useless statement

pilou-:


j00bar:
I'm fine with using `repr()` for using the quotes, however in the case that `tempfile.mkdtemp`is the exception being handled in `__exit__`, `self.temp_dir` won't be defined and the logging statement will raise an Exception.

j00bar:
Thanks for the contribution!

pilou-:
When an exception is raised by `__enter__` method, `__exit__` method isn't called: `__exit__` can not handle an exception raised by `tempfile.mkdtemp`.

pilou-:
@j00bar is that ok with you ?

j00bar:
+1 LGTM thanks!!

PR #282, opened 2016-10-26
Allow to run test suite using tox

pilou-:
This allow to execute the test suite locally using tox.

Once compatibility with Python 3 has been fixed, Python 3 could be added to `.travis.yml` too.

### DIFF
--- a/test/local/ansible/container.yml
+++ b/test/local/ansible/container.yml
@@ -1,7 +1,7 @@
 version: "1"
 services:
-  test:
-    image: python:2.7
+  test-{{ python_version }}:
+    image: "python:{{ python_version }}"
     volumes:
        - "${ANSIBLE_CONTAINER_PATH}:${ANSIBLE_CONTAINER_PATH}" 
        - /var/run/docker.sock:/var/run/docker.sock
--- a/test/local/ansible/main.yml
+++ b/test/local/ansible/main.yml
@@ -1,4 +1,4 @@
-- hosts: test 
+- hosts: test*
   gather_facts: no
   vars:
     project_path: "{{ lookup('env','ANSIBLE_CONTAINER_PATH') }}"
--- a/test/run.sh
+++ b/test/run.sh
@@ -9,11 +9,12 @@
 source_root=$(python -c "from os import path; print(path.abspath(path.join(path.dirname('$0'), '..')))")
 export ANSIBLE_CONTAINER_PATH=${source_root}
 
-image_exists=$(docker images local-test:latest | wc -l)
+export AC_PYTHON_VERSION=$(python -c "import platform; print('.'.join(platform.python_version_tuple()[0:2]))")
+image_exists=$(docker images local-test-"${AC_PYTHON_VERSION}":latest | wc -l)
 if [ "${image_exists}" -le 1 ]; then
    ansible-container --project "${source_root}/test/local" --debug build --local-builder --with-variables ANSIBLE_CONTAINER_PATH="${source_root}"
 fi
 
 ansible-container --project "${source_root}/test/local" --debug run
-status=$(docker inspect --format="{{ .State.ExitCode }}" ansible_test_1)
+status=$(docker inspect --format="{{ .State.ExitCode }}" ansible_test-"${AC_PYTHON_VERSION}"_1)
 exit "${status}"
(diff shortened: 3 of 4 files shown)


chouseknecht:
Thanks for your interest in Ansible Container, and a special thanks for taking the time to submit a PR!

I'm wondering if it's really necessary at this point to consume time and Travis resources to run tests against python 3.4 and 3.5. At some point we will want to begin migrating, but until Ansible moves to 3.x, I don't think there is a rush. 

I'm inclined to merge this and then comment out 3.4 and 3.5 for now.

chouseknecht:
So here's a funny thing... We run the tests inside of a container, which is built on a python:2.7 base image. See [this line](https://github.com/ansible/ansible-container/blob/develop/test/run.sh#L17) in the run.sh script. 

The tests are not actually run using the virtualenv created by *tox*. Of course the `ansible-container build` and `ansible-container run` command would be executed in the virtualenv, but the actual tests end up running in python 2.7 every time.

I think for this to work the way we expect, the local test container needs to use a different base image, and then we need to install and execute tox within that image. I would switch the base image to centos:7, just so we're consistent.

And so then there's no value to running tox outside of the local test container, it just needs to execute within the container.

pilou-:
I added one commit: now the test base image use the same Python version as the one available for the `test/run.sh` script.

The purpose of this PR is only to be able to execute the `ansible-container` in a Python 3 environment _locally_ (not using Travis).

chouseknecht:
Testing this now on my laptop using Docker Machine. First run turned into a blazing dumpster fire. Deleted all containers, and started it again, but same result. Here's the stack trace:

```
ansible_ansible-container_1 exited with code 0
Aborting on container exit...
Stopping ansible_test-2.7_1 ... done
Exporting built containers as images...
Committing image...
Cleaning up temporary directory '/var/folders/g_/mr68c1yn1rl4b7qh6pntbmmm0000gn/T/tmpmms3DY'...
500 Server Error: Internal Server Error ("{"message":"cannot normalise nothing"}")
  File "/Users/chouseknecht/projects/ansible-container/container/cli.py", line 213, in commandline
    getattr(engine, u'cmdrun_{}'.format(args.subcommand))(**vars(args))
  File "/Users/chouseknecht/projects/ansible-container/container/engine.py", line 389, in cmdrun_build
    engine_obj.post_build(host, version, flatten=flatten, purge_last=purge_last)
  File "/Users/chouseknecht/projects/ansible-container/container/docker/engine.py", line 554, in post_build
    for k, v in image_config.items()]
  File "/usr/local/lib/python2.7/site-packages/docker_py-1.8.0-py2.7.egg/docker/utils/decorators.py", line 21, in wrapped
    return f(self, resource_id, *args, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/docker_py-1.8.0-py2.7.egg/docker/api/container.py", line 54, in commit
    json=True)
  File "/usr/local/lib/python2.7/site-packages/docker_py-1.8.0-py2.7.egg/docker/client.py", line 158, in _result
    self._raise_for_status(response)
  File "/usr/local/lib/python2.7/site-packages/docker_py-1.8.0-py2.7.egg/docker/client.py", line 154, in _raise_for_status
    raise errors.APIError(e, response, explanation=explanation)
APIError: 500 Server Error: Internal Server Error ("{"message":"cannot normalise nothing"}")
```

pilou-:
You need to use docker 1.11.2 (same version as the one defined in [./test/local/ansible/main.yml](https://github.com/pilou-/ansible-container/blob/7381da4fadbfa7c94596b27581597254ba61672e/test/local/ansible/main.yml#L32)).

chouseknecht:
I don't think this is quite right. I think what we want is to have pyenv, tox, pyenv-tox and the versions of python to be tested all installed inside the container, and then execute tox inside the container. This would save having to help people through the horror of making all this work on a Mac and other Linux flavors. I still haven't gotten it to work on OSX, and even if I do, I doubt I'll be able to tell anyone how I did it.

chouseknecht:
Working on this now. I'll get it working inside the container. I think we want to change the container to be built on Centos 7 anyway, so that it matches the build container, and I can just do it all in one go.

chouseknecht:
Closing in favor or #297.

Issue #78, opened 2016-06-22
LXD support ?


jpic:
Hi !

I just <3 this project because it opens a wide range of new possibilities.

However, we're not quite satisfied with docker because it's an application containerization system, and we find full system containerization more useful.

Have you considered adding LXD support ?

What's your opinion on it ?

Would you be open for discussion and patch review for that ?

Thanks !

j00bar:
Howdy! We're happy to accept additional container-engine/orchestration modules. Check out `container/engine.py:BaseEngine` for the interface and `container/docker/engine.py` for an implementation. Do note that the interface is not presently stable.

IshwarKanse:
+1 for LXD support. We are using LXD in production at University of Mumbai and it would be really great if ansible-container could support LXD. We are currently building our images manually.

jpic:
I have been a bit busy contributing to lxd_container module for ansible,
and I will be on holidays the next two week, and then will be at europycon.
If anyone wants to start contributing on this go ahead, I will not be
active on this during the next three weeks.

Note that I have some research in novafloss/lxdapi including the
tests/make_busybox script which creates a minimal bootable image in a short
and readable bash script, so that may be useful to whoever wants to get
started on this.

Best regards, to whoever reads this message :)

korotovsky:
We use LXD containers and Ansible for our projects, so +1 for LXD support definitely.

jpic:
I'm just back at work and should start working on this, if anyone wants to
collaborate perhaps we could do a hangouts session or something alike.

dronnix:
+1

jpic:
I have started working on this, let's see how this looks like at the end of the week.

gregdek:
Hey @jpic, any luck here?

jpic:
Hey @gregdek, I think so, basically PR #295 si not complete and some help with internals would be welcome. Interrestingly, the lxd engine seems like it'll be better for a reference implementation than the docker one ie. instead of generating a compose file and calling compose, just make each lxc calls sequentially to build the containers and all, so it's a lot more straightforward. Another interresting major difference with docker is that each running lxd daemon includes image registry features so I would expect this to also help. It would help if a core-developer that has a good vision of ansible-container internals could describe how they would imagine this implementation, what method of the driver should do what exactly, and so on.

jpic:
Sorry for my above response it wasn't of great quality but I wanted to share the idea of working more closely together as soon as I could hehe what I mean is:

[...]

PR #295, opened 2016-10-31
#78: WIP: Lxd

jpic:
This patch currently makes the example to not crash and create lxd containers instead of docker containers. There's still quite some work, but perhaps this draft implementation could be useful to start a discussion about LXD support in ansible-container.


chouseknecht:
@jpic, are you still working on this?

jpic:
I'd like to, but I need more information about the engine, the docker implementation is a bit hard to reverse because it contains a lot more necessary steps than with lxd, so I'm not sure which method should do what exactly, I could make progress with a bit of help.

PR #297, opened 2016-11-02
Use tox and pyenv inside the test container

chouseknecht:
##### ISSUE TYPE
 - Feature Pull Request

##### SUMMARY

Use *tox* inside the testing container and test with Python 2.7, 3.4, and 3.5. 

- created a Centos 7 base image with all the needed packages, shellcheck, pyenv and Python 2.7.12, 3.4.5, and 3.5.0. The image is on Docker Hub as ansible/ansible-container-testing. 
- added test/build-test-img project to use for rebuilding the image
- tests are executed the same as before by running ./test/run.sh
- the same script and process is used locally and in Travis

Current py34 and py35 are commented out in *tox.ini* because there are still syntax issues that need to be fixed. However, py27 completes the tests successfully.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -188,6 +188,7 @@ sftp-config.json
 
 venv/
 *.retry
+.python-version
 
 * Vagrant things
 .vagrant/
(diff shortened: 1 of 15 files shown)


chouseknecht:
This replaces #282

PR #300, opened 2016-11-04
CI: execute the test suite using Python 3.4 and Python 3.5

pilou-:
Another commit with an update of `.mailmap` and `AUTHORS` files.


pilou-:
Note that the Travis builds will take ~20 minutes instead of ~10 minutes (see  #297) : 5 min for the initial setup and then 5 minutes per Python version. Is that a problem ?

chouseknecht:
Do we really need to test with python 3.4? I don't know anything about the differences between 3.4 and 3.5, but it seems like testing with 3.5 would be *close enough*?

@j00bar, thoughts?

pilou-:
Yes, testing with 2.7 and 3.5 seems sufficient: PR rebased and updated (`3.4` removed in `tox.ini`).

pilou-:
Rebased

SalahAdDin:
What's about `python 3.6`?

PR #315, opened 2016-11-18
Use DOCKER_CLIENT_TIMEOUT env var when instantiating Docker client

### LINKED ISSUE #147: Read timeout when creating a large image
I'm creating a large image (~3Gb) and the `ansible-container build` fails at the very end with 

```
Aborting on container exit...
Stopping ansible_jupyterhub_1 ... done
Exporting built containers as images...
Committing image...
UnixHTTPConnectionPool(host='localhost', port=None): Read timed out. (read timeout=60)
```

I tried increasing the timeout using environment variables, as described e.g. [here](https://github.com/docker/compose/issues/1045) but this doesn't seem to have any effect. 

Interestingly, if I do a partial build on the first pass by leaving out some components and bringing the image size below ~ 2Gb, it works. I can then add the other components back in building on top of the previous image and this build finishes fine (now composed of the two layers instead of one). 

versions:

```
Docker version 1.12.0, build 8eab29e
docker-compose version 1.7.0, build 0d7bf73
```

System is mac os x 10.10.5.

chouseknecht:
##### ISSUE TYPE
 - Feature Pull Request

##### SUMMARY
Fixes #147 by using env var DOCKER_CLIENT_TIMEOUT when instantiating the Docker client. If not set, defaults to docker.constants.DEFAULT_TIMEOUT_SECONDS

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -9,6 +9,7 @@ Pierre-Louis Bonicoli <pierre-louis@libregerbil.fr>
 Greg DeKoenigsberg <greg.dekoenigsberg@gmail.com>
 Sandra Wills <swills@ansible.com>
 Dusty Mabe <dusty@dustymabe.com>
+Dylan Silva <thaumos@users.noreply.github.com>
 Charlie Drage <charlie@charliedrage.com>
 Roderick Randolph <roderickrandolph@users.noreply.github.com>
 Todd Barr <tbarr@ansible.com>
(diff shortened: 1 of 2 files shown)


AndydeCleyre:
When prepending an `ansible-container build` command with `DOCKER_CLIENT_TIMEOUT=3000`, the process fails with `Error while fetching server API version: a float is required`, indicating this may need to be casted from a `str`.

chouseknecht:
@AndydeCleyre thanks for pointing this out. #318 addresses this as well as adds support for COMPOSE_HTTP_TIMEOUT. 

I was not able to reproduce the error you're seeing, so if it persists, please open a new issue with more details so we can track it down. 

Thanks!

PR #317, opened 2016-11-19
Change DOCKER_CLIENT_TIMEOUT to COMPOSE_HTTP_TIMEOUT

rinatz:
##### ISSUE TYPE
 - Feature Pull Request

##### SUMMARY
Fixes #315

When using ansible-container with `DOCKER_CLIENT_TIMEOUT` env var, the warning message
`The DOCKER_CLIENT_TIMEOUT environment variable is deprecated.  Please use COMPOSE_HTTP_TIMEOUT instead.` appears on the console.

This message misleads users.
It is simple to use `COMPOSE_HTTP_TIMEOUT` instead of `DOCKER_CLIENT_TIMEOUT`.

### DIFF
--- a/container/docker/engine.py
+++ b/container/docker/engine.py
@@ -744,8 +744,8 @@ class Engine(BaseEngine):
         if not self._client:
             # To ensure version compatibility, we have to generate the kwargs ourselves
             client_kwargs = kwargs_from_env(assert_hostname=False)
-            timeout = os.environ.get('DOCKER_CLIENT_TIMEOUT', DEFAULT_TIMEOUT_SECONDS)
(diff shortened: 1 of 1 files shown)


chouseknecht:
Thanks for taking the time to work on this. However, I think #318 is more complete. It accepts values for both COMPOSE_HTTP_TIMEOUT and DOCKER_CLIENT_TIMEOUT, as probably not everyone is aware that DOCKER_CLIENT_TIMEOUT is being deprecated. It also adds error handling for the unlikely case that int() fails. If you set DOCKER_CLIENT_TIMEOUT=foo, it now fails gracefully.

PR #353, opened 2017-01-10
Brucellino patch typo containery

brucellino:
##### ISSUE TYPE

 - Bugfix Pull Request

##### SUMMARY

There was a simple typo in the error log message, which was bugging me. I tried to find out whether there was some rationale for the use of `containery` instead of `container`, but the internet fell silent. 

This doesn't fix anything that was previously broken, but makes the error message English.


chouseknecht:
LOL. Good catch!

PR #382, opened 2017-02-15
Added option "--use-vault-password-file"

### LINKED ISSUE #381: Support vault password
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Add the capability to use Ansible vault by passing password file.

##### STEPS TO REPRODUCE
- Pass " "--vault-password-file ansible_vault_password.txt" at the end of the ansible-container command (options)
- Place the password file in ansible folder along with main.yml

Will send PR

splisson:
##### ISSUE TYPE
 - Feature Pull Request
##### SUMMARY
Added option "--vault-password-file" that adds : "--vault-password-file ansible_vault_password.txt" to the ansible-playbook call in container/templates/build-docker-compose.j2.yml
The file ansible_vault_password.txt has to exist in ./ansible folder which is mounted as a volume in ansible builder container.

Fixes #381

### DIFF
--- a/container/cli.py
+++ b/container/cli.py
@@ -107,6 +107,8 @@ def subcmd_build_parser(parser, subparser):
     subparser.add_argument('--services', action='store',
                            help=u'Rather than perform an orchestrated build, only build specific services.',
                            nargs='+', dest='service', default=None)
+    subparser.add_argument('--use-vault-password-file', action='store_true',
+                           help=u'Use ansible_vault_password.txt from ansible folder', dest='use_vault_password_file',default=False)
     subparser.add_argument('ansible_options', action='store',
                            help=u'Provide additional commandline arguments to '
                                 u'Ansible in executing your playbook. If you '
--- a/container/templates/build-docker-compose.j2.yml
+++ b/container/templates/build-docker-compose.j2.yml
@@ -2,7 +2,7 @@ ansible-container:
   # If no $DOCKER_HOST then we need to run privileged so we can access /var/run/docker.sock
   {% if not env.DOCKER_HOST %}privileged: true{% endif %}
   image: "{{ builder_img_id }}"
-  command: "/usr/local/bin/builder.sh /usr/bin/ansible-playbook {% if params.debug %}-vvv{% endif %} -i /etc/ansible/ansible-container-inventory.py -c docker {{ params.ansible_options | join(' ')  }} main.yml"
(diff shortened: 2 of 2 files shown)


container/cli.py
```diff
@@ -107,6 +107,8 @@ def subcmd_build_parser(parser, subparser):
     subparser.add_argument('--services', action='store',
                            help=u'Rather than perform an orchestrated build, only build specific services.',
                            nargs='+', dest='service', default=None)
+    subparser.add_argument('--use-vault-password-file', action='store_true',
```
ryansb: This looks like it would have a different behavior from the ansible `vault-password-file` where users can specify the path (to somewhere outside the repo, for example). Do you think the true/false option here covers all the use cases for vault passwords?

splisson: I agree, I admit I took a shortcut while figuring out the relative path and the options. I'm new to python and I need to explore more the arguments parsing. I'll revisit to make it behave like the regular option. The password file would end up inside the builder container  via COPY (instead of volume mount in the current scenario). Any way to avoid files to be committed to the docker image, but just be loaded at runtime ?

chouseknecht:
@splisson, @ryansb  

If I'm not mistaken, you can pass options to the ansible-playbook call using `--` on the command line. See the [reference guide here](http://docs.ansible.com/ansible-container/reference/build.html#cmdoption-arg-ansible-options). 

So the build command would look something like:

```
$ ansible-container build -- --vault-password-file=/path/to/my/vault-password.txt
```

Closing this, because I don't think it's needed. If I'm wrong, will re-open.

splisson:
I'll try that, I think it will work as far as passing the option. The file still needs to be mounted/present in the ansible builder container. If we put the constraint that the file should be in the base_path => /ansible-container/ then that should work. If we want to allow to use a file from somewhere else in the host file system, we still need a small change in builder-docker-compose.j2. yml. I ll test and submit a simple version if needed

chouseknecht:
@splisson, If the file exists outside of the base directory, then you should be able to mount it to the build container using the `--with-volumes` option.

Thanks for understanding why I closed the PR. I appreciate the time you're spending on this, and your willingness to work through the available options. 

Once you get it working the way you want, and assuming the options work out the way we think they should, then a doc update with an example would definitely be welcome. I'm sure others would appreciate learning from your experience. 

Thanks!

splisson:
@chouseknecht I submitted a PR for documentation changes. Thanks for your help

PR #385, opened 2017-02-20
add dockerfile

automaticgiant:
##### ISSUE TYPE
 - Feature Pull Request

##### SUMMARY
add a Dockerfile so you can use from a container

```
#before
$ docker run -it --rm ansible/ansible-container
Unable to find image 'ansible/ansible-container:latest' locally
docker: Error response from daemon: repository ansible/ansible-container not found: does not exist or no pull access.
See 'docker run --help'.
#after
$ docker run -it --rm ansible/ansible-container
usage: ansible-container [-h] [--debug] [--engine ENGINE_NAME]
                         [--project BASE_PATH] [--var-file VAR_FILE]
                         {init,version,run,help,install,push,shipit,stop,restart,build}
                         ...
ansible-container: error: too few arguments
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,4 @@
 from python:2-slim
 run pip install ansible-container
 workdir /work
+cmd ansible-container
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,3 @@
+from python:2-slim
+run pip install ansible-container
+workdir /work
--- a/README.md
+++ b/README.md
@@ -50,6 +50,7 @@ Python sandbox:
 For more details, prerequisite, and instructions on installing the latest development release, please view our 
 [Installation Guide](https://docs.ansible.com/ansible-container/installation.html).
 
+Or, run with `docker run -v "${PWD}:/work" -it --rm ansible/ansible-container`.
 
 ## Getting started
 


j00bar:
Hey, @automaticgiant! Thank you for the contribution and for your interest in the project.

I'd love to see this come into the project, but I've got a couple of thoughts...

1. We should probably be using `ENTRYPOINT` for `ansible-container` and `--help` for `CMD`.
2. Any environment variables for the Docker daemon need to come along too, e.g. `DOCKER_HOST`. If the `DOCKER_HOST` is TLS-secured, you also need to mount a volume for the cert-path. Probably a good idea in the Dockerfile to have things like `ENV DOCKER_HOST $DOCKER_HOST`.
3. (Background: We're currently doing a refactor, followable in the branch `refactor/mk-ii`, which is shaking things up a bit.) I'm a little hesitant to have a Dockerfile in the project-root, as it unintentionally signals that this is the preferred way to run Ansible Container. In Mk. II, Docker isn't even a requirement, so this file might make better sense put into `container/conductor/docker`. @chouseknecht - what are your thoughts on that?

Thanks again, and looking forward to hearing back.

-jag

chouseknecht:
Closing this, as there has been no response since February.

PR #390, opened 2017-03-08
Fixed no message displayed on an Exception.

yanchenko:
##### ISSUE TYPE
 - Bugfix Pull Request

##### SUMMARY
When running  `ansible-container build` on a low-memory machine, it fails with a `MemoryError` with nothing displayed to the user. That's because `str(e)` returns an empty string.
Changed to provide a meaningful output.

Before:
```
Starting Docker Compose engine to build your images...
#
```
After:
```
Starting Docker Compose engine to build your images...
Execution failed with <type 'exceptions.MemoryError'>
#
```

### DIFF
--- a/container/cli.py
+++ b/container/cli.py
@@ -243,5 +243,5 @@ def commandline():
         if args.debug:
             logger.exception(e)
         else:
-            logger.error(e)
+            logger.error('Execution failed with {}'.format(type(e)))
         sys.exit(1)
--- a/container/cli.py
+++ b/container/cli.py
@@ -243,5 +243,6 @@ def commandline():
         if args.debug:
             logger.exception(e)
         else:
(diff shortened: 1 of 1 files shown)


container/cli.py
```diff
@@ -243,5 +243,5 @@ def commandline():
         if args.debug:
             logger.exception(e)
         else:
-            logger.error(e)
+            logger.error('Execution failed with {}'.format(type(e)))
```
chouseknecht: @yanchenko, not sure we always want to see `type(e)`. What we want is a string representation of `e`. Maybe this should check if `str(e)` returns anything, and if not, then return `type(e)`?  I'm concerned that standard exceptions won't result in a text message. Have you tested with a more typical exception?

chouseknecht:
@yanchenko, thanks for taking the extra time on this!

PR #404, opened 2017-03-22
Restructuring modules

j00bar:
Restructuing the codebase so what's outside of the container and what's inside of the container is a clearer line.

1.`container/conductor` moved to `container/lib/conductor`. `lib` has no `__init__.py` so you cannot import `conductor` outside of the container.
2. There's now a `container.common` which has code shared inside and outside of the container. It's referenced as `container.common` no matter whether you load it inside or outside of the conductor
3. We're back to having two engine definitions. One outside. One inside. So `container/engine.py` and `container/docker/engine.py` are back. I made a `container/common/docker.py` which has methods common to them both.

What's still to be done:
1. The various `utils.py` need to be consolidated and cleaned up.
2. The interfaces for `container/engine.py` and `conductor/engine.py` need to be cleaned up.
3. `--devel` is broken.
4. `push` is broken.


j00bar:
We're rejecting this course of action.

PR #414, opened 2017-04-05
Deployment

chouseknecht:
### ISSUE TYPE
 - Feature Pull Request

### SUMMARY
Not completely done, but the engines are capable of generating a playbook, and in the case of `openshift`, I have tested that the playbook runs against `minishift`. 

- Most of the code is being shared. You'll notice a `base_engine.py`, and `base_deploy.py` in the k8s engine path. 
- Installs the ansible.kubernetes-modules role as kubernetes-modules into the conductor
- Installs ansible/module_utils/k8s_common.py into the conductor
- Installs ansible/module_utils/openshift_common.py into the conductor
- Adds k8s engine
- Adds openshift engine
- Adds `settings.k8s_auth` to `container.yml`,  for passing API authentication bits to the Ansible modules
- Adds `settings.k8s_namespace` to `container.yml`, for passing k8s namespace details to the Ansible modules. The namespace can now differ from the Ansible Container project name.
- Fixing bugs I discovered along the way 

#### Generating and running playbooks

After running the `build` command,  generate and run a playbook via the following:

```
$ ansible-container --engine openshift run
```

Adding `--devel` to `run` will cause the playbook to be written to `~/ansible-deployment/playbook.yml`

**Or**

If you just want to generate the playbook:

```
$ ansible-container --engine openshift deploy
```
The playbook is named &lt;project-name&gt;.yml, and by default is writtent o `~/ansible-deployment.`. Use `--output-path` to override the directory where the playbook is written.

#### TODO:
- test the playbook output against `minishift`
- handle persistent volumes. The code from `shipit` has been ported into `k8s/deploy.py`, but it's not quite ready to test.
- after running `deploy` remove the `files` and `templates` folders from the deployment path.
- write integration tests
- documentation

#### Persistent Volume Claims:
The code in `k8s/deploy.py` needs to conform to the new `container.yml` PVC spec that Jag came up with. Same for `openshift/deploy.py`

Here's an example of the new PVC spec:

```
services:
  my-service:
     ...
     volumes:
     - /foo
     -/foo:/bar
     -vol-name:/foo

volumes:
  vol-name:
    docker:
      driver: ...
      driver_opts: ...
      external: ...
    k8s:
      strorage_class: '...'
      access_modes: []
      storage: 10Gi
      selector:
         match_labels: ...
         match_expression: ....
      type: volatile   or  persistent
```


conductor-requirements.txt
```diff
@@ -1,4 +1,5 @@
-git+https://github.com/ansible/ansible.git@v2.3.0.0-0.1.rc1#egg=2.3.3.0
+git+https://github.com/ansible/ansible.git@v2.3.0.0-0.1.rc1#egg=ansible-2.3.0.0
```
ryansb: We can change this over to devel again, now that the `synchronize` bug we were hitting is fixed.

container/core.py
```diff
@@ -32,6 +32,7 @@
 
 REMOVE_HTTP = re.compile('^https?://')
 DEFAULT_CONDUCTOR_BASE = 'centos:7'
+DEFAULT_DEPLOYMENT_PATH = '~/ansible-deployment/'
```
ryansb: I'm thinking this would be better to default to a relative path, basically `os.path.join(config.base_path, 'ansible-deployment/')` how does that sound?

container/cli.py
```diff
@@ -77,6 +77,26 @@ def subcmd_common_parsers(self, parser, subparser, cmd):
                                    help=u'Specify a local path containing roles you want to '
                                         u'use in the Conductor.')
 
+        if cmd in ('deploy', 'push'):
```
ryansb: Is the reason to move these just to avoid duplication between deploy+push, or is there something I'm not seeing?

chouseknecht: Just attempting to avoid duplication.

Issue #437, opened 2017-04-18
Debian based conductor_base fails to build cryptography packages


duffn:
##### ISSUE TYPE
 - Bug Report

##### container.yml 
```
version: "2"
settings:
  # The deployment_output_path is mounted to the Conductor container, so
  # that the `run` and `deployment` commands can write generated Ansible
  # playbooks to it.
  # deployment_output_path: ~/ansible-deployment
  # The Conductor container does the heavy lifting, and provides a portable
  # Python runtime for building your target containers. It should be derived
  # from the same distribution as you're building your target containers with.
  conductor_base: ubuntu:xenial

  # When using the k8s or openshift engines, use the following to authorize with the API.
  # Values set here will be passed to the Ansible modules. Any file paths will be mounted
  # to the conductor container, allowing the `run` command to access the API.
  #k8s_auth:
    # path to a K8s config file
    #config_file:
    # name of a context found within the config.json file
    #context:
    # URL for accessing the K8s API
    #host:
    # An API authentication token
    #api_key:
    # Path to a ca cert file
    #ssl_ca_cert:
    # Path to a cert file
    #cert_file:
    # Path to a key file
    #key_file:
    #boolean, indicating if SSL certs should be validated
    #verify_ssl:

  # When using the k8s or openshift engines, use the following to set the namespace.
  # If not set, the project name will be used. For openshift, the namespace maps to a project,
  # and description and display_name are supported.
  #k8s_namespace:
  #  name:
  #  description:
  #  display_name:

services:
  dwtools:
    from: ubuntu:xenial
    command: ['/bin/false']
```

##### main.yml 
```
- hosts: myhosts
  gather_facts: false
  roles:
    - myrole
```

##### OS / ENVIRONMENT

```
Ansible Container, version 0.9.0.0
Darwin, mng-it-l-060.mycompany.com, 15.6.0, Darwin Kernel Version 15.6.0: Fri Feb 17 10:21:18 PST 2017; root:xnu-3248.60.11.4.1~1/RELEASE_X86_64, x86_64
2.7.10 (default, Oct 23 2015, 19:19:21)
[GCC 4.2.1 Compatible Apple LLVM 7.0.0 (clang-700.0.59.5)] /Users/nicholasduffy/.virtualenvs/ansible-container/bin/python
```
##### SUMMARY
Using a Debian based `conductor_base` fails to build the `pycrypto` and `cryptography` packages.

##### STEPS TO REPRODUCE
[...]

PR #438, opened 2017-04-18
Install gcc when conductor is Debian

chouseknecht:
- Installs gcc when Debian.
- Fixes `config.py` not returning a valid path when `deployment_output_path` = `./ansible-deployment`
- A little clean up in the `init/container.j2.yml` template


chouseknecht:
Fixes #437

PR #452, opened 2017-04-21
Fedora systems do not have or need epel-release

fignew:
##### ISSUE TYPE

 - Bugfix Pull Request

##### SUMMARY

Dockerfile fails to build when using Fedora as distro due to trying to install the unnecessary epel-release package.

```

```


chouseknecht:
@fignew Thanks for the contribution! Will merge once tests complete.

Issue #440, opened 2017-04-18
Error during build on gathering facts with Debian based container


duffn:
##### ISSUE TYPE
 - Bug Report

##### container.yml 
```
version: "2"
settings:
  # The Conductor container does the heavy lifting, and provides a portable
  # Python runtime for building your target containers. It should be derived
  # from the same distribution as you're building your target containers with.
  conductor_base: ubuntu:xenial

  # The deployment_output_path is mounted to the Conductor container, and the 
  # `run` and `deployment` commands then write generated Ansible playbooks to it.
  # deployment_output_path: ./ansible-deployment

  # When using the k8s or openshift engines, use the following to authorize with the API.
  # Values set here will be passed to the Ansible modules. Any file paths will be mounted
  # to the conductor container, allowing the `run` command to access the API.
  #k8s_auth:
    # path to a K8s config file
    #config_file:
    # name of a context found within the config.json file
    #context:
    # URL for accessing the K8s API
    #host:
    # An API authentication token
    #api_key:
    # Path to a ca cert file
    #ssl_ca_cert:
    # Path to a cert file
    #cert_file:
    # Path to a key file
    #key_file:
    #boolean, indicating if SSL certs should be validated
    #verify_ssl:

  # When using the k8s or openshift engines, use the following to set the namespace.
  # If not set, the project name will be used. For openshift, the namespace maps to a project,
  # and description and display_name are supported.
  #k8s_namespace:
  #  name:
  #  description:
  #  display_name:

services: 
  # Add your containers here, specifying the base image you want to build from
  # For example:
  
  web:
    from: ubuntu:xenial
    roles:
      - myrole  
  
registries: {}
  # Add optional registries used for deployment. For example:
  #  google:
  #    url: https://gcr.io
  #    namespace: my-cool-project-xxxxxx   
```

##### main.yml 
The `main.yml` file is not required for >=0.9.0.0.

##### OS / ENVIRONMENT
```
Ansible Container, version 0.9.1rc0
Darwin, mng-it-l-060.mycompany.com, 15.6.0, Darwin Kernel Version 15.6.0: Fri Feb 17 10:21:18 PST 2017; root:xnu-3248.60.11.4.1~1/RELEASE_X86_64, x86_64
2.7.10 (default, Oct 23 2015, 19:19:21)
[...]

PR #453, opened 2017-04-21
Make life better for Xenial users

chouseknecht:
##### ISSUE TYPE
 - Bugfix Pull Request

##### SUMMARY
- Remove `lsb_release` from conductor and force Ansible to read `/etc/lsb-release`
- Replaces `--python-interpreter` with `--use-local-python` option. If True, will prevent mounting `/usr` from conductor to `/_usr` in the target.


chouseknecht:
Fixes #440

PR #460, opened 2017-04-27
Conductor image: support fedora

vrutkovs:
Fedora-based conductor shoud use dnf and install additional package (redhat-rpm-config) -
its required to compile cryptography


chouseknecht:
@vrutkovs Thanks for taking the time to sort out the Fedora conductor build. Merging.

PR #495, opened 2017-05-04
Updates docs and starter container.yml delivered by init to match each other.

### LINKED ISSUE #493: Update init default container.yml to match docs
##### ISSUE TYPE

 - Feature Idea

##### SUMMARY

Currently, running `ansible-container init` populates $PWD with a `container.yml` that's different from the one described in the Getting Started Guide's Dipping a Toe In - Starting from Scratch section. Updating `container.j2.yml` to match the docs would round off a sharp corner that might trip up some users (like yours truly).

wbrefvem:
##### ISSUE TYPE

 - Docs Pull Request

##### SUMMARY

Fixes #493 

The new `container.yml` delivered by `ansible-container init` matches the one in the docs and is buildable and runnable.

### DIFF
--- a/container/templates/init/container.j2.yml
+++ b/container/templates/init/container.j2.yml
@@ -41,17 +41,16 @@ settings:
 services: {}
   # Add your containers here, specifying the base image you want to build from.
   # To use this example, uncomment it and delete the curly braces after services key.
+  # You may need to run `docker pull ubuntu:trusty` for this to work.
 
-  # version: "2"
-  # services:
-  #   web:
-  #     from: "ubuntu:trusty"
-  #     ports:
-  #       - "80:80"
-  #     command: ["/usr/bin/dumb-init", "/usr/sbin/apache2ctl", "-D", "FOREGROUND"]
-  #     dev_overrides:
-  #       environment:
-  #         - "DEBUG=1"
+  # web:
+  #   from: "ubuntu:trusty"
+  #   ports:
+  #     - "80:80"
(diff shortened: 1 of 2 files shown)


container/templates/init/container.j2.yml
```diff
@@ -39,20 +39,19 @@ settings:
...
-  #     environment:
-  #       - "DEBUG=1"
-  #
+  # Add your containers here, specifying the base image you want to build from.
+  # To use this example, uncomment it and delete the curly braces after services key.
+
+  # version: "2"
+  # services:
+  #   web:
+  #     from: "ubuntu:trusty"
```
evgeni: shouldn't this still be centos:7? :)

wbrefvem: This is copypasta from the [Getting Started Guide](http://docs.ansible.com/ansible-container/getting_started.html). The idea is that the example service produced by init should be identical to the one in the docs. Whether or not they both should be centos is something I have no particular opinion on, as long as the UX is smooth and intuitive.

evgeni: Right. I stand corrected ;)

wbrefvem: Just realized i duplicated the services key. I'll push a fix.

wbrefvem: Update: trying to build the ubuntu:trusty image from a centos conductor is non-starter. I've updated the docs and `container.j2.yml` to use centos. So I guess I do have an opinion after all. :)

PR #501, opened 2017-05-04
Making tests work again in 0.9.1

aaronthebaron:
##### ISSUE TYPE

 - Bugfix Pull Request

##### SUMMARY

The test are not setup to work at all in 0.9.0. 

Added a line to pull the ansible/ansible-container-testing image if it's not there, since that wasn't happening.

It seems like there is still a dependency problem in docker-py which I have a fix in for as well. docker/docker-py/pull/1581

Alas, the tests are still failing because of the error below but they are closer to working.

```
Error: No such object: ansible_test_1
```


chouseknecht:
@aaronthebaron 

Thanks for taking the time to work on tests!

So, we're actually trying to move tests into roles. To execute tests, we run `python ./setup.py test`, which executes the [run_tests.yml playbook](https://github.com/ansible/ansible-container/blob/develop/test/run_tests.yml), which in turn executes roles found in `tests/roles`.

Where we could use help is possibly migrating the old tests to this new format. However, since there was so much code refactor for the 0.9 release, I'm not sure how much of it can be migrated, or even makes sense.

--Chris

aaronthebaron:
No problem. I see where you're going with that. Feel free to close this, I thought I would at least make as far as I got available to you. Thanks!

chouseknecht:
@aaronthebaron, thanks! Closing for now.

PR #521, opened 2017-05-15
Enable image pulls, and fix role install

chouseknecht:
##### ISSUE TYPE
 - Bugfix Pull Request

##### SUMMARY
- If a service does not have a `roles` directive, then pass the `from` value as the `image` to Ansible playbook that runs the application, and allow the `docker_service` module to pull the image.
- Warns that it will attempt to pull the image
- Fixes parsing of `meta/container.yml`.
- Adds a new service with the same name as the role, since the meta data no longer contains a service name.
- Only adds a `roles` directive to the new service, ignoring all other directives

### DIFF
--- a/container/docker/engine.py
+++ b/container/docker/engine.py
@@ -514,21 +514,25 @@ class Engine(BaseEngine):
 
         service_def = {}
         for service_name, service in self.services.items():
+            service_definition = {}
             if service.get('roles'):
                 image = self.get_latest_image_for_service(service_name)
                 if image is None:
                     raise exceptions.AnsibleContainerConductorException(
                         u"No image found for service {}, make sure you've run `ansible-container build`".format(service_name)
                     )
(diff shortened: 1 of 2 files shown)


container/docker/engine.py
```diff
@@ -514,21 +514,26 @@ def generate_orchestration_playbook(self, url=None, namespace=None, local_images
...
-                image = self.client.images.get(service['from'])
-                if image is None:
-                    raise exceptions.AnsibleContainerConductorException(
-                        u"No image found for service {}. You probably need to pull the image from a repository.".format(service_name)
-                    )
-            service_definition = {
-                u'image': image.tags[0],
-            }
+                try:
+                  image = self.client.images.get(service['from'])
```
menghan: @chouseknecht here the indent is 2 instead of 4

chouseknecht: @menghan, thanks for catching. Fixed in #523.

PR #548, opened 2017-05-24
Split --no-cache into two options

### LINKED ISSUE #482: Role caching is too eager
##### ISSUE TYPE
 - Bug Report

##### container.yml 
```
version: "2"
settings:
  conductor_base: debian:8.6

services:
  eager-cache:
    from: debian:8.6
    roles:
      - common

registries: {}
```

##### roles/common/tasks/main.yml
```
- shell: echo $(date) > /buildstamp
```

##### OS / ENVIRONMENT
```
Ansible Container, version 0.9.1rc0
...
```

##### SUMMARY
`get_role_fingerprint` only takes the content of the role into account. While the example above is a little contrived it, this is a real issue when doing updates using the package manager for example. There you expect different results even though the role itself has not changed at all.

##### STEPS TO REPRODUCE
run `ansible-container build` twice

##### EXPECTED RESULTS
Two images with different values in /buildstamp

##### ACTUAL RESULTS
Only one image is produced

##### Discussion
I think that keeping the role caching behaviour would be fine if there was a way to disable the cache without also disabling the conductor cache, as done by --no-cache. I think it makes sense to split --no-cache into --no-conductor-cache and --no-(container|service|?)-cache.

ddelbondio:
##### ISSUE TYPE
 - Feature Pull Request

##### SUMMARY
Two new options (--no-conductor-cache and --no-container-cache) are
added to allow fine grained caching control. The old --no-cache will
be kept and be an "alias" for which sets both new options.

Fixes #482

### DIFF
--- a/container/cli.py
+++ b/container/cli.py
@@ -128,11 +128,20 @@ class HostCommand(object):
                                help=u'Rather than build all services, only build specific services.',
                                nargs='+', dest='services_to_build', default=None)
         subparser.add_argument('--no-cache', action='store_false',
+                               help=u'Shortcut for --no-conductor-cache and --no-container-cache.',
+                               dest='cache', default=True)
+        subparser.add_argument('--no-conductor-cache', action='store_false',
+                               help=u'Ansible Container caches conductor images during builds '
+                                    u'and reuses the conductor image if it determines no '
+                                    u'changes have been made necessitating rebuild. '
+                                    u'You may disable conductor caching with this flag.',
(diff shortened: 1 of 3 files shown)


j00bar:
👍

chouseknecht:
@ddelbondio 

Thanks for working on this!  Can you please rebase, and we'll get it merged.

ddelbondio:
Done

PR #568, opened 2017-05-31
Adds repository_prefix registry attribute

### LINKED ISSUE #550: Offer a way to override the repository name from container.yml
##### ISSUE TYPE

 - Feature Idea

##### SUMMARY

Here is a little bit of background on what I am trying to achieve. We are using [Jenkins Pipelines](https://jenkins.io/doc/book/pipeline/) to do CI/CD with ansible-container. Jenkins pipelines have the concept of [multibranch pipeline](https://jenkins.io/doc/book/pipeline/multibranch/) which mean that the project can keep track of multiple branches and build the ansible-container project from a specific branch when a change occurs.

Thanks to the `--project-name` option recently introduced in ansible-container we are now able to run concurrent build of the same ansible-container project without having any issue. The only issue we have is with registry names. Our docker registry is hosted in AWS ECR. When we change `--project-name` attribute, we have to create new registries as the `--project-name` is used in the repository path when we use the push command.

For example our ansible-container project is named `myapp`. This project manages 3 different services `builder`, `nginx`, `nodejs`. As a result of that, We already need to create 3 different repositories in AWS ECR:

* mynamespace/myapp-builder
* mynamespace/myapp-nginx
* mynamespace/myapp-nodejs

Let's say that I want to keep track of 3 type of branches on my CI/CD server `master`, `stage`, `dev`. We will name those environment `env`.

The Jenkins server can now continuously build in parallel the `myapp` project if I run ansible-container build --project-name `myapp-<env>`. But now it means that I need to maintain 9 AWS ECR repositories for a single project... which is not sustainable to me especially if you want to keep track of all type of branches to spin up ephemeral environments on AWS ECR clusters for developers.

* mynamespace/myapp-master-builder
* mynamespace/myapp-master-nginx
* mynamespace/myapp-master-nodejs
* mynamespace/myapp-stage-builder
* mynamespace/myapp-stage-nginx
* mynamespace/myapp-stage-nodejs
* mynamespace/myapp-dev-builder
* mynamespace/myapp-dev-nginx
* mynamespace/myapp-dev-nodejs

##### SUGGESTION:

It could be awesome to add an extra parameter in the `registries` key from the `container.yml` so that we can lock down the repositories to a specific value like:

```
### myapp/container.yml
registries:
   aws:
     url: 123456789.dkr.ecr.us-east-1.amazonaws.com
     namespace: mynamespace
     registry_name: myapp  ### TBD
```

Thanks to that, no matter what project-name I define for `myapp` I know all images will be pushed to the same repository and I can then use custom tagging to differentiate them. It also makes image promotion way easier (Let's say `stage` passes integration test and get signed by QA it can be promoted to `master` just by re-applying a tag to the image).

Please let me know what you think about this and if you need extra info, let me know !
Thanks!

chouseknecht:
Closes #550 

##### ISSUE TYPE
 - Feature Pull Request

##### SUMMARY
Allows overriding the repository name prefix. Today we push images to a repository consisting of <project_name>-<service_name>. This change provides a mechanism for overriding the <project_name> portion of the repository name.

### DIFF
--- a/container/core.py
+++ b/container/core.py
@@ -324,11 +324,13 @@ def push_images(base_path, image_namespace, engine_obj, config, **kwargs):
     registry_name = engine_obj.default_registry_name
     namespace = image_namespace
     save_conductor = config.get('settings', {}).get('save_conductor_container', False)
+    repository_prefix = None
 
     if push_to:
         if config.get('registries', dict()).get(push_to):
             url = config['registries'][push_to].get('url')
             namespace = config['registries'][push_to].get('namespace', namespace)
+            repository_prefix = config['registries'][push_to].get('repository_prefix')
             if not url:
                 raise AnsibleContainerRegistryAttributeException(
                     u"Registry {} missing required attribute 'url'".format(push_to)
@@ -338,6 +340,8 @@ def push_images(base_path, image_namespace, engine_obj, config, **kwargs):
 
     if username and not password:
         # If a username was supplied without a password, prompt for it
+        if url != engine_obj.default_registry_url:
+            registry_name = url
         while not password:
             password = getpass.getpass(u"Enter password for {0} at {1}: ".format(username, registry_name))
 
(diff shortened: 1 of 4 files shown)


Lowess:
I can confirm it works like a charm!

PR #576, opened 2017-06-03
get volumes defined, mounted at container build time

kkoralsky:
##### ISSUE TYPE

 - Bugfix Pull Request

##### SUMMARY

Volumes defined in container.yaml should be available also during container build phase. I belive this was intended by the authors, since in 668-678 of container/core.py there is loop over those volumes and it works if someone decides to use container's local python.

### DIFF
--- a/container/core.py
+++ b/container/core.py
@@ -694,7 +694,7 @@ def conductorcmd_build(engine_name, project_name, services, cache=True,
                         pass
 
                     # Use the conductor's Python runtime
-                    run_kwargs['volumes'] = {engine.get_runtime_volume_id('/usr'): {'bind': '/_usr', 'mode': 'ro'}}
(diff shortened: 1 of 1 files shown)


chouseknecht:
@kkoralsky 

Thank you for taking the time to submit this PR.

@j00bar, I didn't realize volumes weren't being mounted to the target container due to a bug. Thought we were purposefully not mounting volumes during build.

We've talked about this before. We said we didn't want to mount volumes to the target because it leaves `Mounts` and `Volumes` in the image meta data. However, we're mounting`/_usr` most of the time, and that ends up in the meta data already.

To me, not being able to mount volumes makes the build process more complicated. What we really need is a way to modify the image meta data post build. 

I'm inclined to merge this.

chouseknecht:
@kkoralsky 

I agree that we need a way to bind volumes during the build. It's annoying and inefficient to copy things back and forth between the conductor and the targets, so I'm inclined to merge this.

PR #579, opened 2017-06-03
Fix compilation error when running under Python 3.5

LeHack:
##### ISSUE TYPE
 - Bugfix Pull Request

##### SUMMARY
This compilation error was the only obstacle I encountered when running ansible-container with Python 3.5, despite having no issues at all with Docker and Ansible standalone.
I tested the full build/run/deploy cycle and everything seems to work fine with this fix applied.

### DIFF
--- a/container/docker/engine.py
+++ b/container/docker/engine.py
@@ -23,7 +23,7 @@ import tarfile
 try:
     import httplib as StatusCodes
 except ImportError:
-    from http import HTTPStatus as StatusCodes
+    import http.client as StatusCodes
 
 import container
 from container import host_only, conductor_only


j00bar:
Not being a heavy user of Python3 yet, I'm a little confused. I read this in the docs: https://docs.python.org/3/library/http.html#http.HTTPStatus -- by that read, the code as present is correct. Could you help me understand your PR? Thanks!

LeHack:
Correction, this concerns Python 3.4 (but my fix works also with 3.5).
Actually regarding Python I'm not much more than a novice.
I found this suggestion [somewhere on stackoverflow](https://stackoverflow.com/questions/13778252/import-httplib-importerror-no-module-named-httplib) while trying to resolve the issue that popped up during the build process:
```
$ ansible-container build 
ERROR	Unknown exception	
  File "/home/.../testapp/lib/python3.4/site-packages/container/docker/engine.py", line 23, in <module>
    import httplib as StatusCodes
ImportError: No module named 'httplib'

During handling of the above exception, another exception occurred:

  File "/home/.../testapp/lib/python3.4/site-packages/container/cli.py", line 268, in __call__
    getattr(core, u'hostcmd_{}'.format(args.subcommand))(**vars(args))
  File "/home/.../testapp/lib/python3.4/site-packages/container/__init__.py", line 28, in __wrapped__
    return fn(*args, **kwargs)
  File "/home/.../testapp/lib/python3.4/site-packages/container/core.py", line 137, in hostcmd_build
    config['services'], **kwargs)
  File "/home/.../testapp/lib/python3.4/site-packages/container/utils/loader.py", line 16, in load_engine
    package='container')
  File "/home/.../testapp/lib/python3.4/importlib/__init__.py", line 109, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 2254, in _gcd_import
  File "<frozen importlib._bootstrap>", line 2237, in _find_and_load
  File "<frozen importlib._bootstrap>", line 2226, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 1200, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 1129, in _exec
  File "<frozen importlib._bootstrap>", line 1471, in exec_module
  File "<frozen importlib._bootstrap>", line 321, in _call_with_frames_removed
  File "/home/.../testapp/lib/python3.4/site-packages/container/docker/engine.py", line 25, in <module>
    from http import HTTPStatus as StatusCodes
ImportError: cannot import name 'HTTPStatus'
```

ekohl:
I can confirm that the current develop branch is broken on python 3.4. Like https://docs.python.org/3/library/http.html#http.HTTPStatus mentions it is new in 3.5. I don't think the current code is really the best solution since ```http.client``` is a different package from the enum ```HTTPStatus``` and importing it this way might lead to subtle bugs.

I'm installing 3.5 and 3.6 to check if I can come with a recommendation for a proper fix.

ekohl:
Having looked into it I'd recommend the following:

```python
try:  # Python 3.5+
    from http import HTTPStatus as StatusCodes
except ImportError:
    try:  # Python 3
        from http import client as StatusCodes
    except ImportError:  # Python 2
        import httplib as StatusCodes
```

chouseknecht:
@LeHack 

Would you mind modifying this to use the method suggested by @ekohl. Once done, I can merge this. 

Thank you for your efforts on this, and apologies for not looking at it in quite awhile.

rinatz:
I'm in trouble because of this error.
Can I fix it instead of @LeHack?

5444de:
when i input "ansible-container build",show this error:
image

5444de:
i have a error when i first test ansible-container，why：
image

PR #613, opened 2017-06-19
Sanity checks for directories on import, import --force option

josiahg:
##### ISSUE TYPE

 - Feature Pull Request

##### SUMMARY

Adding sanity checking for directories on import. 
Check for the presence of `Dockerfile` in the target directory, and warn if this is present.
Check for the presence of the `roles/` directory or `container.yml`, warn if this is present and offer the `--force` option to overwite.
Implement the `import --force` option.

The force option works well in most cases, but currently fails with an error if directories are part of an import. `shutil.copytree` does not support overwrite. We could add code to remove directories if force is enabled, but this may have undesired impact.

Please provide feedback on the preferred approach in the above situation. Also appreciate any feedback on general code style and functionality. Not sure about the exception names I chose, for example.

Let me know and I'll be glad to update.

```

```


chouseknecht:
@josiahg 

This looks good! Thanks for working on it. 

My inclination around fixing the issue with `shutil.copytree` would be to call `os.listdir()` on `self.import_from`, and remove any matching files found in `target`. Since the user chose the `--force` option, it's expected that we will be replacing these items.

container/core.py
```diff
@@ -431,7 +431,9 @@ def hostcmd_version(base_path, project_name, engine_name, **kwargs):
 
 
 @host_only
-def hostcmd_import(base_path, project_name, engine_name, **kwargs):
+def hostcmd_import(base_path, project_name, engine_name, force=False, **kwargs):
+    kwargs['force'] = force
```
chouseknecht: Since we're not actually using `force` here, we don't need to expose it in the parameters, and therefore don't need this line. Just accept that it's part of `kwargs`, and allow it to be passed along.

chouseknecht:
@josiahg 

Looks perfect! 

One minor nit. Don't punch me in the face. :-)

After that adjustment, we'll merge it.

PR #628, opened 2017-06-29
Add Stateful Sets proposal

ehelms:
##### ISSUE TYPE
 - Proposal Pull Request

##### SUMMARY
Proposal for ansible-container to support Stateful Sets in k8s and OpenShift environments. I figured I'd start with a proposal this time instead of code.This is supported on k8s 1.5+ and already exists within the Ansible Kubernetes modules. I believe this should be fairly simple for me to implement but I think supporting this and the syntax are worth a discussion.

### DIFF
--- a/proposals/statefulsets.md
+++ b/proposals/statefulsets.md
@@ -1,0 +1,75 @@
+# Adding Stateful Sets Support
+
+## Objectives
+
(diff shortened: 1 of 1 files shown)


chouseknecht:
@ehelms 

Looks pretty straightforward to me. I say, go for it.

fire:
Ping! Wanted to show there's interest!

Issue #225, opened 2016-09-09
Support rkt as a local container engine


j00bar:

srizzling:
Whats required for something like this to kickstart? 0.4 milestone is pretty close :P

TomasTomecek:
@srizzling I think that biggest issue will be that there is no rkt-compose, so the orchestration for rkt needs to be done from scratch.

srizzling:
Yeah, I was kind of implying that is the milestone 0.4 is very tight for a big task like this. 

rkt also doesn't support the full lifecycle that docker supports either (commit, push and pull).

PR #639, opened 2017-07-06
Add proposal for Multiple Containers per Pod

marc-sensenich:
##### ISSUE TYPE

 - Proposal Pull Request

##### SUMMARY

Proposal that allows for multiple containers per pod to build services in container.yml that can use container patterns such as sidecars and ambassador. Currently a [TODO in the code](https://github.com/ansible/ansible-container/blob/0d54b0efba8a6bc205c0619226397201381e8e20/container/k8s/base_deploy.py#L358)

### DIFF
--- a/proposals/multiple-containers.md
+++ b/proposals/multiple-containers.md
@@ -1,0 +1,50 @@
+# Supporting Multiple Containers Per Service in Kubernetes and OpenShift
+
+## Table of Contents
+- [Reasoning](#Reasoning)
+- [Declaring Multiple Containers Per Service](#declaring-multiple-containers-per-service)
+
+## Reasoning
+
+Some services are composed of multiple containers that are coupled and share resources to for a single cohesive service. Common cases of this are sidecar and ambassador containers. Within Ansible Container currently it is a single pod per service, which allows for the generation of the images used within these services separately and deployed to a server. However, the container.yml file is not able to act as a single source-controlled piece of code to define the state of the deployable application. Therefore, I propose allowing multiple containers to be declared per service. 
+
+## Declaring Multiple Containers Per Service
+
+To achieve this the service section will accept a new property, containers, which will allow for the declaration of 1-N containers per service. The addition of this property will allow for the maintenance of existing applications by using an check when building and deploying the service; e.g. `if 'containers' in service`. A declaration may look like this:
+
+```yaml
+services:
+  web:
+    containers:
+      - from: centos:7
+        entrypoint: [/usr/bin/entrypoint.sh]
+        working_dir: /
+        user: apache
+        command: [/usr/bin/dumb-init, httpd, -DFOREGROUND]
+        ports:
+          - 8000:8080
+          - 4443:8443
+        roles:
+          - apache-container
+        volumes:
+          - static-content:/var/www/static
+      - from: centos:7
+        command: [/usr/bin/dumb-init, file-refresher, --out-dir, /var/www/static]
+        entrypoint: [/usr/bin/entrypoint.sh]
+        user: refresher
+        roles:
+          - apache-file-refresher
+        volumes:
+          - static-content:/var/www/static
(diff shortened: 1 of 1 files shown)


proposals/multiple-containers.md
```diff
@@ -0,0 +1,50 @@
+# Supporting Multiple Containers Per Service in Kubernetes and OpenShift
+
+## Table of Contents
+- [Reasoning](#Reasoning)
+- [Declaring Multiple Containers Per Service](#declaring-multiple-containers-per-service)
+
+## Reasoning
+
+Some services are composed of multiple containers that are coupled and share resources to for a single cohesive service. Common cases of this are sidecar and ambassador containers. Within Ansible Container currently it is a single pod per service, which allows for the generation of the images used within these services separately and deployed to a server. However, the container.yml file is not able to act as a single source-controlled piece of code to define the state of the deployable application. Therefore, I propose allowing multiple containers to be declared per service.
```
ehelms: Minor comment, you could drop the last sentence here since this will end up in the repository as a proposal for the project itself.

ehelms:
Howdy @marc-sensenich , thanks for the proposal! I think it would be helpful to have some examples or descriptions of how ths setup would be handled in the Docker and Kubernetes case. For example, including what the resulting Kubernetes template would look like and how Docker would handle this when it can have a single service name per container.

marc-sensenich:
Thanks for reviewing it @ehelms ! I added some descriptions regarding how Docker could manage a list of containers per service, given that it would only be able to map one container per service. Let me know if there are any more updates needed or your thoughts.

chouseknecht:
@marc-sensenich 

Thanks for submitting this. I'm running into this issue myself. I need to add a sidecar for database access, and not sure how to do it in Ansible Container. Started looking at your proposal, and contemplating how to approach the problem. Still mulling it over.

In the meantime, added your proposal to the agenda for the next containers WG. See [#225 here](https://github.com/ansible/community/issues/225). If you can attend, and help with the discussion, it would be much appreciated.

marc-sensenich:
@chouseknecht I should be able to attend the next WG to be a part of the discussions.

chouseknecht:
Accepted during the 8/14 container meeting

awiddersheim:
This will be an awesome addition for sure. One thing I think is missing in this proposal is how networking will work with multiple containers in a "pod". 

In a k8s/openshift pod each container shares the same network namespace. If I have a Redis container running in the same pod as my application, my application's configuration would point to Redis at `localhost:6379` since the network namespace is shared.

In the case of Docker or something like docker-compose where each container is it's own service you would refer to the Redis instance differently in the application's config, usually by service name and Docker's DNS service discovery will point you to the right place. For example, `redis:6379`.

You can make docker-compose behave more like k8s pods and have services share the same network namespace. It's a little funky but I've gotten it to work with something like this:

```
version: "2.1"

services:
  # NOTE(awiddersheim): This container is meant to be a basis for some
  # of the other containers networking. All of the ports are published
  # under this service and other containers will adopt it's network
  # namespace.
  network:
    image: alpine:latest
    ports:
      - 127.0.0.1:8443:8443
    command: ["tail", "-f", "/dev/null"]
    stop_signal: SIGKILL

  redis:
    image: redis:4.0-alpine
    network_mode: "service:network"

  some_rest_app:
    image: some_rest_app:latest
    network_mode: "service:network"
```

I guess my concern here is the examples work fine with things like volumes but when it comes to networking pieces the configurations you may use in your templates and things may differ based on where things run.

tima:
@awiddersheim: Thanks for your feedback. I think I follow you. If you have a proposal, especially something that's not as "funky", I think we'd all love to hear it. There's always room for improvement.

awiddersheim:
@tima Sure. I don't have any well thought out solution. Just noticed this as a possible pain point as work happens in #685.

PR #642, opened 2017-07-06
Added capability to set Downward API Environment Variables

marc-sensenich:
Added a separate directive to the service to allow for the addition of Downward API values into the running containers

##### ISSUE TYPE

 - Feature Pull Request

##### SUMMARY

Adds an additional option on a service which allows for the use of `downward_api` when using Kubernetes or OpenShift to expose Downward API values into the container.


chouseknecht:
@marc-sensenich 

Thank your for taking the time to work on this. Much appreciated!

Would you be able to add some doc updates explaining when and how to use this feature. Thanks!

PR #674, opened 2017-08-01
Update split in docker pull

Demonsthere:
##### ISSUE TYPE
 - Bugfix Pull Request

##### SUMMARY
When pulling an image from a private registry this error occurs:

```python

--
File "/usr/bin/conductor", line 11, in <module>
load_entry_point('ansible-container', 'console_scripts', 'conductor')()
File "/_ansible/container/__init__.py", line 19, in __wrapped__
return fn(*args, **kwargs)
File "/_ansible/container/cli.py", line 392, in conductor_commandline
**params)
File "/_ansible/container/__init__.py", line 19, in __wrapped__
return fn(*args, **kwargs)
File "/_ansible/container/core.py", line 659, in conductorcmd_build
cur_image_id = engine.pull_image_by_tag(service['from'])
File "/_ansible/container/__init__.py", line 19, in __wrapped__
return fn(*args, **kwargs)
File "/_ansible/container/docker/engine.py", line 560, in pull_image_by_tag
repo, tag = image.split(':')
ValueError: too many values to unpack
```

This is due to the fact that the image is in the format `registry_url:port/dir/image:tag`, which has two semicolons, and the current split will not work correctly. 

This fixes this issue by using rsplit and splitting on the first right-foremost semicolon

### DIFF
--- a/container/docker/engine.py
+++ b/container/docker/engine.py
@@ -557,7 +557,7 @@ class Engine(BaseEngine, DockerSecretsMixin):
         repo = image
         tag = 'latest'
         if ':' in image:
-            repo, tag = image.split(':')
+            repo, tag = image.rsplit(':',1)
         logger.debug("Pulling image {}:{}".format(repo, tag))
         try:
             image_id = self.client.images.pull(repo, tag=tag)


chouseknecht:
@Demonsthere Thanks for fixing this!

PR #685, opened 2017-08-09
Multi-container implementation

chouseknecht:
##### ISSUE TYPE
 - Feature Pull Request

##### SUMMARY
Implements support for multiple containers per pod as proposed by #639

### DIFF
--- a/container/k8s/base_deploy.py
+++ b/container/k8s/base_deploy.py
@@ -370,6 +370,14 @@ class K8sBaseDeploy(object):
                     container[key] = value
             return container, volumes
 
+        def _update_volumes(existing_volumes, new_volumes):
+            existing_names = {}
+            for vol in existing_volumes:
+                existing_names[vol['name']] = 1
+            for vol in new_volumes:
+                if vol['name'] not in existing_names:
+                    existing_volumes.append(vol)
+
         templates = CommentedSeq()
         for name, service_config in iteritems(self._services):
             containers = []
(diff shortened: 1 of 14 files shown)


chouseknecht:
Just need to add documentation, and this is ready to go.

container/k8s/base_deploy.py
```diff
@@ -323,13 +346,10 @@ def _service_to_container(name, service):
                     if vols:
                         volumes += vols
```
marc-sensenich: This can lead to duplicate volumes if two containers are mounting the same volume

container/k8s/base_deploy.py
```diff
@@ -348,24 +368,31 @@ def _service_to_container(name, service):
...
+        for name, service_config in iteritems(self._services):
+            containers = []
+            volumes = []
+            pod = {}
+            if service_config.get('containers'):
+                for c in service_config['containers']:
+                    cname = "{}-{}".format(name, c['container_name'])
+                    k8s_container, k8s_volumes, = _service_to_k8s_container(name, c, container_name=cname)
+                    containers.append(k8s_container)
+                    volumes += k8s_volumes
```
marc-sensenich: This can lead to duplicate volumes if two containers are mounting the same volume
```
spec.template.spec.volumes[1].name: Duplicate value: \"static-content\""
```

chouseknecht: Makes sense. I'll see if I can fix.

PR #698, opened 2017-08-12
Add CLI options to specify /src mount path and volume driver

marcusianlevine:
##### ISSUE TYPE

 - Feature Pull Request

##### SUMMARY

I'm attempting to setup Docker-out-of-Docker (i.e. mounting `/var/run/docker.sock`) in my Dockerized Jenkins CI environment, and found that (because Docker volume mounts via the socket always refer to host paths) I could not get the working directory (which is inside a Docker container) to mount into the sibling conductor container as `/src`.

To solve this, I need to mount `/src` with the path to my Jenkins `$JENKINS_HOME/workspace` on the **host** rather than the path to the workspace on the Jenkins build container.

This pull request adds an optional CLI flag to the `build` step, which overrides `base_path` in `engine.py` when mounting `/src` with the provided path

If the flag is not provided, `base_path` is mounted as usual.

I also added a flag to support the same Jenkins workflow, which supports mounting named volumes managed outside of Docker local host volume driver.

Also I caught another broken `self` reference in the k8s engine...

##### Proposed syntax

```
ansible-container build --src-mount-path /path/to/project/on/docker/host

ansilbe-container build --volume-driver rancher-nfs
```

### DIFF
--- a/container/cli.py
+++ b/container/cli.py
@@ -150,6 +150,10 @@ class HostCommand(object):
                                     u'into target containers in order to run Ansible. Use when the target '
                                     u'already has an installed Python runtime.',
                                dest='local_python', default=False)
+        subparser.add_argument('--src-mount-path', action='store',
+                               help=u'Specify the host path that should be mounted to the conductor at /src.'
+                                    u'Defaults to the directory from which ansible-container was invoked.',
+                               dest='src_mount_path', default=None)
         subparser.add_argument('ansible_options', action='store',
                                help=u'Provide additional commandline arguments to '
                                     u'Ansible in executing your playbook. If you '
--- a/container/cli.py
+++ b/container/cli.py
@@ -71,6 +71,10 @@ class HostCommand(object):
                                    help=u'Mount one or more volumes to the Conductor. '
                                         u'Specify volumes as strings using the Docker volume format.',
                                    default=[])
+            subparser.add_argument('--volume-driver', action='store',
+                                   help=u'Specify volume driver to use when mounting named volumes '
+                                        u'to the Conductor.',
+                                   default=None)
             subparser.add_argument('--with-variables', '-e', action='store', nargs='+',
                                    help=u'Define one or more environment variables in the '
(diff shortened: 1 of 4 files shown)


marcusianlevine:
I ended up switching to Docker-in-Docker instead of DooD since the host mounting problem was still giving me trouble

Going to leave this open since I think these options do work and are useful in very specific use-cases

chouseknecht:
@marcusianlevine 

Thank your for working on this!

Looks good to me. Would you mind updating the reference docs with the new CLI options.

marcusianlevine:
Sure, I think I put that in the right place...

chouseknecht:
I read your doc update, and realized that I misunderstood the use case. I *thought* you were attempting to mount the `base_path` to something other than `/src` in the conductor. That *seemed* like something that other user might want to do as well. 

However, that's obviously not with this does. What I *think* it does is allow you to run the build command from path `/foo` while mounting some other path, `/bar` to the conductor at `/src`. That seems strange, and confusing to me, but I'm sure it fits your use case. I'm not a Jenkins expert, nor a containerized Jenkins expert. 

Now that I sorta understand, maybe the parameter should be `--src-path`, and not contain the word `mount`. To me that feels slightly less confusing.

@j00bar, any thoughts on this?

j00bar:
So I have no particular thoughts or opinions on the particular use case, but I will say that the option is probably one that's overall beneficial to the project. We've always encouraged the use case where `container.yml` and other artifacts are at the root of the `project_path` that is the `/src` used during the build process. We've gotten a non-trivial number of issues, mailing list posts, and IRC questions from folks who put their `container.yml` and other build artifacts in a subdirectory of what they want to be the `/src` for the build. I still firmly believe that `container.yml` in the base of the project should be the strongly encouraged practice -- it ensures that a newcomer to a project can see it and identify it as a project buildable with Ansible Container. But it would provide utility to our user-base to have that behavior be overridable, and in such a way that the mild inconvenience still connotes the preferred practice of keeping `container.yml` et al in the project root.

Does that make sense?

Issue #399, opened 2017-03-16
It is not possible to override the 'command' directive to start the container with


dmsimard:
##### ISSUE TYPE
 - Bug Report

##### container.yml 
```
version: "2"
services:
  test:
    image: jrei/systemd-centos:latest
    command: /usr/sbin/init
    volumes:
      - /sys/fs/cgroup:/sys/fs/cgroup:ro
```

##### main.yml 
```
- hosts: test
  gather_facts: false
  tasks:
    - command: systemctl daemon-reload
```

##### OS / ENVIRONMENT
```
$ ansible-container --debug version
Ansible Container, version 0.2.0
Linux, hostname, 4.9.13-200.fc25.x86_64, #1 SMP Mon Feb 27 16:48:42 UTC 2017, x86_64
2.7.13 (default, Jan 12 2017, 17:59:37) 
[GCC 6.3.1 20161221 (Red Hat 6.3.1-1)] /tmp/ansible-container/bin/python
No DOCKER_HOST environment variable found. Assuming UNIX socket at /var/run/docker.sock
{u'Architecture': u'x86_64',
 u'BridgeNfIp6tables': True,
 u'BridgeNfIptables': True,
 u'CPUSet': True,
 u'CPUShares': True,
 u'CgroupDriver': u'cgroupfs',
 u'ClusterAdvertise': u'',
 u'ClusterStore': u'',
 u'Containers': 2,
 u'ContainersPaused': 0,
 u'ContainersRunning': 0,
 u'ContainersStopped': 2,
 u'CpuCfsPeriod': True,
 u'CpuCfsQuota': True,
 u'Debug': False,
 u'DefaultRuntime': u'runc',
 u'DockerRootDir': u'/var/lib/docker',
 u'Driver': u'devicemapper',
 u'DriverStatus': [[u'Pool Name', u'docker-253:0-1445263-pool'],
                   [u'Pool Blocksize', u'65.54 kB'],
                   [u'Base Device Size', u'10.74 GB'],
                   [u'Backing Filesystem', u'xfs'],
                   [u'Data file', u'/dev/loop0'],
                   [u'Metadata file', u'/dev/loop1'],
                   [u'Data Space Used', u'3.66 GB'],
                   [u'Data Space Total', u'107.4 GB'],
                   [u'Data Space Available', u'23.97 GB'],
                   [u'Metadata Space Used', u'5.308 MB'],
                   [u'Metadata Space Total', u'2.147 GB'],
[...]

PR #719, opened 2017-09-08
Added support for override command directive when conductor is building image

balll:
##### ISSUE TYPE
 - Feature Pull Request

##### SUMMARY

Added support for override command directive when conductor is building image with systemd as pid 1. (feature idea  #399)

Default command when conductor use to run docker when building image is

`sh -c "while true; do sleep 1; done`

This PR will allow user to override it with `build_command` in each service.

```
version: 2

settings:
  project_name: test
  conductor:
    base: centos:7

services:
  grafana:
    from: centos/systemd
    roles:
      - grafana
    volumes:
      - /sys/fs/cgroup:/sys/fs/cgroup:ro
    stop_signal: SIGRTMIN+3
    cap_add: ["SYS_ADMIN"]
    command: /usr/sbin/init
    tty: true
    build_command: /usr/sbin/init
```

### DIFF
--- a/container/core.py
+++ b/container/core.py
@@ -731,6 +731,9 @@ def conductorcmd_build(engine_name, project_name, services, cache=True, local_py
                     environment=dict(ANSIBLE_CONTAINER=1)
                 )
 
+                if service.get('build_command'):
+                    run_kwargs['command'] = service['build_command']
+
                 if service.get('volumes'):
                     for volume in service['volumes']:
                         pieces = volume.split(':')


chouseknecht:
@balll 

Thank you for taking the time to work on this. We always appreciate feedback and contribution, so a big "Thank You!"

If we decide to accept this, I suspect it needs a bit more documentation, including an example. We may also want to adjust how it's implemented. 

I would like to discuss this on Monday at the next [Container WG meeting](https://github.com/ansible/community/tree/master/group-container). I'll put it on the agenda. I *think* the meeting time is 11:00 PM in Thailand, so there is no expectation that you attend. I believe that you have provided sufficient detail for the group to discuss and decide next steps.

balll:
@chouseknecht Thank you for your response. I've updated an example in the first comment.

chouseknecht:
@balll 

The container working group decided the following:

1. Your use case is sound, and it's one we should support
2. Rather than `build_command` as a new directive, we would prefer `build_overrides`
3. The new `build_overrides` directive will be similar to `dev_overrides`, except it will apply to `build` rather than `run`.  Any settings in `build_overrides` will take precedence when the `build` command is invoked. 
4. If no `build_overrides` is present in `container.yml`, then `build` will function just as it does today, overriding `command` with `sh -c "while true; do sleep 1; done`.

Using this new approach, your example `container.yml` from above would become the following:

```
version: 2

settings:
  project_name: test
  conductor:
    base: centos:7

services:
  grafana:
    from: centos/systemd
    roles:
      - grafana
    volumes:
      - /sys/fs/cgroup:/sys/fs/cgroup:ro
    stop_signal: SIGRTMIN+3
    cap_add: ["SYS_ADMIN"]
    command: /usr/sbin/init
    tty: true
    build_overrides:
      command: /usr/sbin/init
```

balll:
@chouseknecht I've pushed changes reflexed container working group decision.

TheoNolasco:
+1 

Ran into this issue trying to install Percona on CentOS 7.  Realized theres no way for my to startup MySQL properly.. Unless I'm missing something?

container/core.py
```diff
@@ -731,6 +731,13 @@ def conductorcmd_build(engine_name, project_name, services, cache=True, local_py
                     environment=dict(ANSIBLE_CONTAINER=1)
                 )
 
+                if service.get('build_overrides'):
```
chouseknecht, line 734: I think this actually belongs in `config.py`, and it should be handled as part of the [set_env method](https://github.com/balll/ansible-container/blob/9e2a80a88864a4f66f51283d33905283d10c7735/container/config.py#L116) on the host side. That method would then accept a new `env`  value of `build`, and return the appropriate configuration.

danielkza:
Is this till being worked on? Being able to build systemd containers is fundamental for me to start using ansible-container. Any development/testing help needed?

balll:
@danielkza When I have some free time, will work on it. Or submitting your own PR is fine.

KernelPryanic:
@chouseknecht Sorry, I'm new here, but why do we need separate `env` for this functionality?

sambanks:
Has this been abandoned? We also need this functionality to use the tool and would be happy to lend a hand.

Voronenko:
@sambanks You can try to issue PR based on current develop

sambanks:
Thanks @voronenko but I have a workaround (manually linking into etc/systemd/system/multi-user.target.wants/) which has kept us moving for now.

Voronenko:
Well, if you know solution, perhaps formulate it into some article (saying medium) and refer here.  We might add some KB articles to Wiki than.

Idea of the opensource products are - if you can improve, improved locally, etc - return your commitment to community.

sambanks:
It's not a solution, it's just a workaround for our specific use case. If you manually link the service file, as I shared in the previous message, it will start on the next boot, but it's very far from proper systemd support.

haneul:
it would be really nice if it's supported.

PR #723, opened 2017-09-10
Updated all instances of the text conductor-container...

TheoNolasco:
... to be container-conductor. The conductor_base refers the use of "container-conductor" and images being pulled to generate the conductor from Ansible use the same.

##### ISSUE TYPE
 - Docs Pull Request

##### SUMMARY
 - The images being pulled to generate a default conductor use the: `tag = 'container-conductor-%s:%s' % (base_image.replace(':', '-')`, which are not properly reflected in the documents as well as the args when prompting ansible-container to "Leave the Ansible Builder Container intact upon build completion"

- Changes as follows: 

docs/rst/conductor.rst
42. It should be named ``container-conductor-$DISTRO-$TAG:$ANSIBLE_CONTAINER_VERSION``,
44. 1.0rc1, your Conductor base image should be named ``container-conductor-opensuse-42.3:1.0rc1``.

docs/rst/reference/build.rst
25. option:: --save-container-conductor

container/cli.py
131. subparser.add_argument('--save-container-conductor', action='store_true',

### DIFF
--- a/container/cli.py
+++ b/container/cli.py
@@ -128,7 +128,7 @@ class HostCommand(object):
                                     u'previously built image for your hosts. Disable '
                                     u'that with this flag.',
                                dest='purge_last', default=True)
-        subparser.add_argument('--save-container-conductor', action='store_true',
+        subparser.add_argument('--save-conductor-container', action='store_true',
                                help=u'Leave the Ansible Builder Container intact upon build completion. '
                                     u'Use for debugging and testing.', default=False)
         subparser.add_argument('--services', action='store',
--- a/container/cli.py
+++ b/container/cli.py
@@ -128,7 +128,7 @@ class HostCommand(object):
                                     u'previously built image for your hosts. Disable '
                                     u'that with this flag.',
                                dest='purge_last', default=True)
-        subparser.add_argument('--save-conductor-container', action='store_true',
+        subparser.add_argument('--save-container-conductor', action='store_true',
                                help=u'Leave the Ansible Builder Container intact upon build completion. '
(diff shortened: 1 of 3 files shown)


chouseknecht:
@TheoNolasco 

We appreciate your attention to this issue, and we value your feedback. 

I do have a question though... 

Please explain the need to change `--save-conductor-container` to `--save-container-conductor`. It's not related to the image names, and the phrase "conductor container" makes sense. I think this would be a breaking change.

TheoNolasco:
@chouseknecht - Thanks for the reply!

My thought of adjusting the parameter `--save-conductor-container` to `--save-container-conductor` was consistency, tbh. This might initially confuse someone if they knew the naming conventions being used of the conductor image (ie, the image name is 'container-conductor-%s:%s', why is the command to save it conductor-container). 

Though if this parameter could potentially break something else, or confuse someone even more, I can definitely re-submit without this adjustment. Thanks again!

j00bar:
Oh, man. That's a tricky one. :)

The idea behind the conductor container's name is that you pull from Docker Hub `ansible/container-conductor-${distro}-${version}`, as in the Conductor image for Ansible Container. The commandline parameter is for the Conductor container in Ansible Container.

Perhaps we're being too clever with the `ansible/container` part. Opinions welcome.

TheoNolasco:
@j00bar - Definitely makes sense. 

In my rush, I thought I could just re-adjust the change back and commit... when I realize I should have just reverted. Ugh. I've decided to create a new PR with just the docs changes. Thanks for the explanation.

PR #737, opened 2017-09-21
Handle case where role/meta/main.yml is an empty file.

maverickagm:
##### ISSUE TYPE
 - Bugfix Pull Request

##### SUMMARY
If a role contains an empty meta/main.yml file ansible-container errors with a "None has to attribute get" error. This is because yaml.safe_load() can return None but this case is not handled. This pull request handles that case.


chouseknecht:
@maverickagm 

Thanks! We appreciate the contribution. Looks good to me.

PR #790, opened 2017-11-13
add support for buildah backend

TomasTomecek:
##### ISSUE TYPE
 - Feature Pull Request

##### SUMMARY
This PR adds full buildah support.

TODO:
- [ ] use ~~`kpod`~~ podman to pull images
- [x] try to store config.json inside conductor
- [ ] rename to OCI
- [ ] tests
- [ ] docs
- [ ] RFE: ansible wants to write to ~/.ansible -- tell it to write to /tmp so that whole container can be R/O
- [ ] `s/kpod/[podman](https://koji.fedoraproject.org/koji/packageinfo?packageID=26289)/`


TomasTomecek:
I took the recently contributed hello-world project from @j00bar and successfully built it using buildah:
```
$ buildah images
ccb0b0bd7c05         docker.io/library/hello-world-conductor:latest           Feb 10, 2018 17:32     878.2 MB
0d451e4a7a7c         docker.io/library/hello-world-flask:20180210173302       Nov 14, 2017 21:07     423.4 MB
0d451e4a7a7c         docker.io/library/hello-world-flask:latest               Nov 14, 2017 21:07     423.4 MB
```

(don't mind those weird dates)


I had to tinker the role and `container.yml` a bit:
 * `s/synchronize/copy/`
 * `s/yum/dnf/`
 * `s/centos:7/fedora:27` -- I actually had to create base conductor on my own
   * `conductor_base: docker.io/library/container-conductor-fedora-27:0.9.3rc0`

This PR is very far from being complete or even useful: docker engine implementation in a-c is so beefy that it will take quite some time to port it to buildah.

TomasTomecek:
I'm closing this PR as I no longer have time to work on it. Please carry it if you have the resources.

jchristi:
Buildah and Ansible Container sound like very similar tools. How does buildah compare with ansible container? Where do they overlap and diverge? ELI5 :smiley:

j00bar:
If you think of a `docker build`, it's really two parts - the part that manages instantiating a build filesystem to be changed and committing the changes with metadata; and the part that interprets a Dockerfile as build instructions. Buildah is the first half - offering a daemonless way to manage the build of a container image without having an opinion on how the changes get actually made. The idea with this engine is that those changes would be made with Ansible, and buildah would offer a Dockerless way of doing the rest.

jchristi:
Having little knowledge beforehand about Ansible Container, I think the important piece of information I was missing (after having read through the [docs](https://docs.ansible.com/ansible-container/getting_started.html#building-the-service)), is that Ansible Container actually runs a docker daemon when building an image:
```
Copying build context into Conductor container.
Docker™ daemon integration engine loaded. Build starting.       project=hello-world
Building service...     project=hello-world service=flask
```
So the feature this PR would provide is the ability to swap out the docker daemon used at build time for buildah instead, which is daemonless.

j00bar:
Ansible Container is designed to lean on plugable "Engine" implementations for each of its major workflow steps. The `docker` engine is (currently) the default, and an engine need not implement all capabilities. The `docker` engine is currently the only one that supports build. The Buildah engine would also support build, but wouldn't (for example) support run or deploy.

jchristi:
Per @TomasTomecek's blog post: [Building Container Images with Buildah and Ansible](https://blog.tomecek.net/post/building-containers-with-buildah-and-ansible/), it appears that if all I want to do is build a container with Ansible and Buildah, I can use vanilla Ansible with the [buildah connection plugin](https://docs.ansible.com/ansible/latest/plugins/connection.html#plugin-list) without having to use Ansible Container. If I also wanted container pushing, deploying and orchestration though, then I would need Ansible Container. Does that sound correct?

PR #794, opened 2017-11-17
Jag/reuse build artifacts

j00bar:
##### ISSUE TYPE
 - Feature Pull Request

##### SUMMARY
In rebuilds, when the cache gets "busted", we apply a role cleanly from the previous image. To speed up rebuilds, it makes sense to reapply the role to the same previous intermediate container, in order to take advantage of the performance gains from Ansible's idempotence. This pull request implements such functionality.

Also, this pull request cleans up a ton of test cruft.


j00bar:
Discussion: do we implement a command-line flag to specifically disable this feature? Or do we say you use this full cache strategy or no cache strategy at all?

chouseknecht:
@j00bar 

I say we merge it, and let user feedback determine which command line flags are needed, if any.

PR #804, opened 2017-11-25
extra_hosts now works on docker

mdiez-modus:
##### ISSUE TYPE
 - Bugfix Pull Request

##### SUMMARY
Ansible-Container was ignoring the `extra_hosts` setting from `container.yml`.

Fortunatelly it was an easy fix: all I had to do was to pass this setting inside of 

`container/docker/engine.py generate_orchestration_playbook()`

So now all my containers have `extra_hosts` !


mdiez-modus:
wrong account...

PR #805, opened 2017-11-25
extra_hosts now works on docker

marcosdiez:
##### ISSUE TYPE
 - Bugfix Pull Request

##### SUMMARY
Ansible-Container was ignoring the `extra_hosts` setting from `container.yml`.

Fortunatelly it was an easy fix: all I had to do was to pass this setting inside of 

`container/docker/engine.py generate_orchestration_playbook()`

So now all my containers have `extra_hosts` !

### DIFF
--- a/container/docker/engine.py
+++ b/container/docker/engine.py
@@ -760,6 +760,9 @@ class Engine(BaseEngine, DockerSecretsMixin):
                         service_definition['volumes'] = []
                     service_definition['volumes'].append("{}:/run/secrets:ro".format(self.secrets_volume_name))
 
+            if 'extra_hosts' in service:
+                service_definition['extra_hosts'] = service['extra_hosts']
+
             logger.debug(u'Adding new service to definition',
                          service=service_name, definition=service_definition)
             service_def[service_name] = service_definition
--- a/container/docker/engine.py
+++ b/container/docker/engine.py
@@ -141,7 +141,7 @@ class Engine(BaseEngine, DockerSecretsMixin):
     COMPOSE_WHITELIST = (
         'links', 'depends_on', 'cap_add', 'cap_drop', 'command', 'devices',
         'dns', 'dns_opt', 'tmpfs', 'entrypoint', 'environment', 'expose',
-        'external_links', 'labels', 'links', 'logging', 'log_opt', 'networks',
+        'external_links', 'extra_hosts', 'labels', 'links', 'logging', 'log_opt', 'networks',
         'network_mode', 'pids_limit', 'ports', 'security_opt', 'stop_grace_period',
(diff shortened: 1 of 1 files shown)


chouseknecht:
@marcosdiez 

Thanks for the contribution! We appreciate you taking the time to work on this. 

On the surface, this doesn't feel right to me. We don't want to enumerate every Docker compose directive during orchestration. At one point there was a _whitelist_ of supported directives, which existed because we were limiting directives to those that translated easily to K8s.

I'll research this a bit more, and see if I can provide a better response. 

Thanks for your patience!

marcosdiez:
Hi @chouseknecht 

extra_hosts is documented.
That's why I added...

https://docs.ansible.com/ansible-container/container_yml/reference.html#services

https://docs.ansible.com/ansible-container/container_yml/reference.html#extra-hosts

marcosdiez:
Now.... it's funny that you don't plan to implement all docker-compose feature.

In my current project, I have the following setup and I am pretty happy with it:

- all containers are made using ansible containers
- after they are done, I run a python script that converts `container.yml` into `docker-compose.yml`

that guarantees that my developers and my cloud machines run exactly the same configuration.

I planned, after finishing this project, to push back the code to convert `container.yml` do `docker-compose.yml`  
(sorry, waiting one minute to launching containers is too much for us)

but now that you are telling me that, by design you want to to offer as little settings as possible, I am kind of unsure if I picked the right tool for the job.

container/docker/engine.py
```diff
@@ -760,6 +760,9 @@ def generate_orchestration_playbook(self, url=None, namespace=None, vault_files=
                         service_definition['volumes'] = []
                     service_definition['volumes'].append("{}:/run/secrets:ro".format(self.secrets_volume_name))
 
+            if 'extra_hosts' in service:
```
chouseknecht: Add `extra_hosts` to COMPOSE_WHITELIST above, and this should just work.

chouseknecht (changes requested):
See comment

chouseknecht:
@marcosdiez 

I think your point regarding `waiting one minute to launching containers is too much for us` is completely valid. Rather than forcing the docker engine to orchestrate via playbook, I wonder if we should go back to orchestrating via the `docker` Python module.

mdiez-modus:
Sorry, @chouseknecht 
now I understand what you meant.

I thought there was some documentation about which commands you guys plan to implement because because the aim of the tool would be to be a common denominator between docker and k8s.

About `waiting one minute to launching containers is too much for us`, well... less waiting is better than more waiting :)

chouseknecht (approved):

PR #850, opened 2018-01-09
Deployment artifacts proposal

chouseknecht:
##### ISSUE TYPE
 - Feature Pull Request

##### SUMMARY

Adds proposal to change the artifacts produced by `deploy`. 

Rather than producing a monolithic playbook, produce a better organized solution that relies on an Ansible role and uses the new K8s modules.

### DIFF
--- a/proposals/deployment-artifiacts.md
+++ b/proposals/deployment-artifiacts.md
@@ -1,0 +1,50 @@
+# Refactor Deployment Artifacts
+
+## Summary
+
+Produce a portable, better organized deployment artifact that uses an Ansible role, and the new Ansible 2.5 K8s modules.
+
+## Details
+
+Deployment currently generates the following artifacts:
+
+- Single, monolithic playbook containing all tasks, and all K8s object configurations inline
+- Copy of the `ansible.kubernetes-modules` role
+
+This proposal is to replace the current artifacts with the following:
+
+- Playbook that executes a role
+- Role containing K8s object configuration files separate from playbook tasks
+
+The generated role will include the following:
+
+- Task files that group tasks by function (e.g., start, stop, destroy, etc.)
+- A separate configuration file for each K8s object
+- README.md file for the role, providing how-to information, and contents descriptions
+
+Additional details include:
+
+- The playbook name and role name will match the project name
+- Tasks will be tagged, so that the playbook can be used to perform an action based on tag, as it does today
+- New modules available in Ansible 2.5 will be used, rather than the `ansible.kubernetes-modules` role.
+
+Artifacts will be written to the following directory structure:
+
+```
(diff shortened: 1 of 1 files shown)


proposals/deployment-artifiacts.md
```diff
@@ -0,0 +1,50 @@
...
+Additional details include:
+
+- The playbook name and role name will match the project name
+- Tasks will be tagged, so that the playbook can be used to perform an action based on tag, as it does today
+- New modules available in Ansible 2.5 will be used, rather than the `ansible.kubernetes-modules` role.
+
+Artifacts will be written to the following directory structure:
+
+```
+ansible-deployment/
```
TomasTomecek, line 34: where will the role(s) be placed?

chouseknecht: @TomasTomecek 

Was thinking we would continue writing artifacts into the `ansible-deployment` subdirectory that `deploy` adds to the project.

TomasTomecek: I was just curious where would the generated role(s) be placed. You wrote that

    Produce a portable, better organized deployment artifact that uses an Ansible role...

When I look at the proposed directory hierarchy, I'm not sure which directory will have the resulting, generated role(s).

Also, what will be the default action of the root playbook `project-name.yml`?

proposals/deployment-artifiacts.md
```diff
@@ -0,0 +1,50 @@
...
+  project-name.yml
+  project-name/
```
chouseknecht, line 36: @TomasTomecek 

This was my thinking for the role. It would just be written out in it's own directory, rather than placing it under a `/roles` subdirectory. I should call this directory `project-name-role` to be more clear.

proposals/deployment-artifiacts.md
chouseknecht: I'm thinking the playbook is really simple. Maybe something like the following:

```
- name: my-project
  hosts: all
  gather_facts: no
  roles:
    - role: my-project-role
      tags:
      - always
```

proposals/deployment-artifiacts.md
TomasTomecek: oh right, makes perfect sense now; feel free to leave it as it is

TomasTomecek (approved):
Good proposal!

ehelms:
This might be leaping ahead, and more of "phase 2" but potentially good for the proposal itself. I do find the various tags useful for large operations. But I also find myself wanting to more easily re-build and re-deploy a single or set of services rather than the entire stack. If the role design could somehow facilitate that, through tags or variables that would allow faster development, testing and deployment cycles.

PR #853, opened 2018-01-20
conductor: add F27, drop F24

TomasTomecek:


TomasTomecek:
good catch with the tests!

PR #909, opened 2018-03-30
replace characters that are unsafe in docker layer names with a "_"

jimrobinson:
##### ISSUE TYPE
 - Bugfix Pull Request

##### SUMMARY
If ansible-container has a path to a role directory that is nested (e.g., roles/some/path/to/role/) it ends up trying to use a layer name that is invalid in Docker.

Docker layer names may not contain characters outside the range [a-zA-Z0-9_.-].

This patch replaces all characters outside the allowed range with underscores.

### DIFF
--- a/container/core.py
+++ b/container/core.py
@@ -696,7 +696,8 @@ def _find_base_image_id(engine, service_name, service):
     return image_id
 
 def _intermediate_build_container_name(engine, service_name, image_fingerprint, role_name):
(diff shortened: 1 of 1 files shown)


Voronenko:
@jimrobinson your fix is in develop now.

Voronenko:
@jimrobinson fix is in develop as a part of cumulative PR last week, thanks for your input!

PR #920, opened 2018-04-16
Fix pip 10 parse_requirements

### LINKED ISSUE #919: Support for pip 10 - breaking changes introduced.
- Bug Report
##### SUMMARY

Pip has now has reached stable version 10, which has breaking change causing 
ansible-container to fail during install

```
    ImportError: No module named req
```

##### STEPS TO REPRODUCE

Ensure you have latest stable pip 
```
  pip install -U pip
```

Try to install ansible-container

```
pip install ansible-container[docker]
```
##### EXPECTED RESULTS
ansible-container gets installed

##### ACTUAL RESULTS

Installation fails with stacktrace final lines:

```
    ImportError: No module named req
```

tobill:
##### ISSUE TYPE
 - Bugfix Pull Request

##### SUMMARY

Fixes #919 

```

```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,7 +7,10 @@ import distutils.log
 from setuptools import setup, find_packages
 from setuptools.command.test import test as TestCommand
 from setuptools.command.sdist import sdist as SDistCommand
-from pip.req import parse_requirements
+try: # for pip >= 10
+    from pip._internal.req import parse_requirements
+except ImportError: # for pip <= 9.0.3
+    from pip.req import parse_requirements
 import container
 
 class PlaybookAsTests(TestCommand):


marcosdiez:
This works perfectly over here. You somebody else is having this problem (Ubuntu 16.04) for example, just install ansible-container like this: 

`pip install git+https://github.com/tobill/ansible-container.git@develop`

cikupin:
need this PR to be merged into master in order to support pip 10 <span class="emoji-outer emoji-sizer"><span class="emoji-inner" style="background: url(chrome-extension://immhpnclomdloikkpcefncmfgjbkojmh/emoji-data/sheet_apple_32.png);background-position:30.0235017626322% 4.054054054054054%;background-size:5418.75% 5418.75%" data-codepoints="1f44d"></span></span>

marcusianlevine (approved):
The build is failing for what seems like an unrelated reason, but this at least gets the build further than where it currently fails on ansible/develop and gets `pip install` working

marcusianlevine:
@chouseknecht @j00bar can we get this merged? We have three open PRs now addressing the same pip regression #926 #923 

We need to address the otherwise failing tests but `ansible:develop` doesn't even get past installation...

gregdek:
Fixed by #939 instead. Thanks!

PR #921, opened 2018-04-18
Make pip install command ZSH-compatbile in README

### LINKED ISSUE #902: In the case of macOS the installation method is different
### ISSUE TYPE

* Documentation Report

## Installation
"https://github.com/ansible/ansible-container#installing"

`pip install ansible-container[docker,openshift]`

 but on MacOS you can install only with `pip install ansible-container docker openshift`.

Is this knowledge correct?

By the way, when you run the installation on MacOS it looks like the following.

```
 %pip install ansible-container[docker,openshift]
zsh: no matches found: ansible-container[docker,openshift]
```

attilanagy:
##### ISSUE TYPE
 - Docs Pull Request

##### SUMMARY
The pip installation commands in the README file doesn't work for users using ZSH as square brackets have a special meaning. The easiest way to escape them is to use quotes in the command line.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,14 +34,14 @@ Use Ansible Container to manage the container lifecycle from development, throug
 
 Install using *pip*, the Python package manager:
 
-    $ sudo pip install ansible-container[docker,openshift]
+    $ sudo pip install "ansible-container[docker,openshift]"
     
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @attilanagy!

PR #926, opened 2018-05-10
Fix import of parse_requirements in setup.py for pip>=10

krisfris:
##### ISSUE TYPE

 - Bugfix Pull Request

##### SUMMARY
Pip 10 was released a few weeks ago and moved parse_requirements from pip.req to
pip._internal.req. Thus, setup of ansible-container with pip 10 fails with this error:

```
      File "<string>", line 1, in <module>
      File "/tmp/pip-install-q5c29n85/ansible-container/setup.py", line 10, in <module>
    ImportError: No module named 'pip.req'
```

I applied the fix described here https://stackoverflow.com/a/49867265/7662112

```

```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,7 +7,12 @@ import distutils.log
 from setuptools import setup, find_packages
 from setuptools.command.test import test as TestCommand
 from setuptools.command.sdist import sdist as SDistCommand
-from pip.req import parse_requirements
+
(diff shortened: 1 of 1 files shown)


marcusianlevine:
Thanks for submitting but this is a duplicate of #920

krisfris:
Oh I see

PR #939, opened 2018-06-05
Cumulative tests fix - lets get project great again (at least to green tests state)

Voronenko:
Colleagues, lets leave develop branch in stable state with green tests at least.
It is too nice project to leave it in current "middle" state.
 
cc: @j00bar , @gregdek , @chouseknecht

##### ISSUE TYPE
 - Bugfix Pull Request

##### SUMMARY
1. Fixed travis tests to green state - we have a deadly red line for a while
2. safe characters filtering for role name in edge cases 
3. Amazon:2 tests - added tar which is now absent in base image
4. Workaround for non-english file names if any
5. Pinned docker version to prevent breakage for 3.x.x changes
6. Support for PIP 10

Changes are UP and green:

https://travis-ci.com/softasap/ansible-container/builds/75429794

PR #923, opened 2018-04-30
Allow to run setup with pip < 9.0.3

rpelisse:
##### ISSUE TYPE
 - Bugfix Pull Request

##### SUMMARY
I'm so python developer, so feel free to closed and never merge this. I've just run into this issue, trying to play with Ansible Container on my Fedora 27, so I though I should share the fix.

PR #941, opened 2018-06-07
Updates setup.py to work with pip9 and pip10

### LINKED ISSUE #940: Ansible container is not compatible with Pip10
More or less:

$  pip install --upgrade pip
$  pip install ansible-container[docker,k8s]

brianbruggeman:
Fixes #940 

[pip10 intentionally broke compatibility](https://github.com/pypa/pip/issues/5154#issuecomment-378197307) to prevent using internals.
However, there is no particular package that actually provides a way to
read requirements and convert them into a setup.py file, so package
developers are stuck with using pip.*.req.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,8 +7,14 @@ import distutils.log
 from setuptools import setup, find_packages
 from setuptools.command.test import test as TestCommand
 from setuptools.command.sdist import sdist as SDistCommand
-from pip.req import parse_requirements
 import container
+
+try:
+    # pip10
+    from pip._internal.req import parse_requirements
+except ImportError:
+    # pip9
+    from pip.req import parse_requirements
 
 class PlaybookAsTests(TestCommand):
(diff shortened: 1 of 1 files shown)


Voronenko:
Also addressed in:

https://github.com/ansible/ansible-container/pull/920

https://github.com/ansible/ansible-container/pull/923/

https://github.com/ansible/ansible-container/pull/931/

https://github.com/ansible/ansible-container/pull/939

Let's support one single pull request instead?

brianbruggeman:
> Let's support one single pull request instead?

I'm cool with that.  I did a search on Pip 10 and I was surprised that didn't see anything in the results.  

It's such a stupid easy fix, though, that it would take 10 minutes to review and add...  why hasn't it been added in the past month?  Additionally, it makes me wonder if the project is dead/dying without that level of support... because this type of failure is just an automatic no-go.

gregdek:
Thanks @brianbruggeman -- we merged @Voronenko PR #939 for this one.

PR #947, opened 2018-07-14
fix(tests) https validation in older python

Voronenko:
##### ISSUE TYPE
 - Bugfix Pull Request

##### SUMMARY
Fixes https issues with pip during conductor image baking in wheezy, precise, trusty


j00bar:
Probably a good idea when you do `pip install` to use `--no-cache-dir` to keep image sizes down.

Voronenko:
True,  https://github.com/ansible/ansible-container/pull/948/files

Issue #944, opened 2018-06-25
Destroy command deletes all similar images of other projects too


matteotanca-sardex:
- Bug Report
##### container.yml 

```
version: "2"
settings:

  conductor:
    base: ubuntu:16.04
    environment:  # List or mapping of environment variables
      - DISTRO=xenial

  project_name: site1

services:
  web:
    from: ubuntu:16.04
    roles:
      - apache2
      - sardex/site1-web
    depends_on:
      - "mysql_database"
    ports:
      - "127.0.0.1:80:80"
      - "127.0.0.1:443:443"
    volumes:
      - "$PWD/sources/site1:/var/www/html:rw"
    entrypoint: ["/etc/init.d/startup.sh"]
    command: ["start"]

  mysql_database:
    from: ubuntu:16.04
    roles:
      - mysql
      - sardex/site1-db
    ports:
      - "127.0.0.1:3306:3306"
    entrypoint: ["/etc/init.d/startup.sh"]
    command: ["start"]

registries: {}

```
```
version: "2"
settings:

  conductor:
    base: ubuntu:16.04
    environment:  # List or mapping of environment variables
      - DISTRO=xenial

  project_name: site2
  
services:
  web:
    from: ubuntu:16.04
    roles:
      - apache2
      - sardex/site2-web
    depends_on:
      - "mysql_database"
    ports:
      - "127.0.0.1:80:80"
      - "127.0.0.1:443:443"
    volumes:
      - "$PWD/sources/site2:/var/www/html:rw"
    entrypoint: ["/etc/init.d/startup.sh"]
    command: ["start"]

  mysql_database:
    from: ubuntu:16.04
    roles:
      - mysql
      - sardex/site2-db
    ports:
      - "127.0.0.1:3306:3306"
    entrypoint: ["/etc/init.d/startup.sh"]
    command: ["start"]

registries: {}

```
##### OS / ENVIRONMENT

```
bug version
Ansible Container, version 0.9.2
Linux, matteo-pc, 4.13.0-45-generic, #50~16.04.1-Ubuntu SMP Wed May 30 11:18:27 UTC 2018, x86_64
3.6.5 (default, May  3 2018, 10:08:28) 
[GCC 5.4.0 20160609] /home/matteo/.virtualenvs/sites-53tickPf/bin/python3.6

```
##### SUMMARY
After building those two container.yml, if I destroy one project, both projects conductors are deleted so I have to rebuild both.

##### STEPS TO REPRODUCE

[...]

PR #950, opened 2018-07-15
fix(docker) remove images that relate to current project only

Voronenko:
##### ISSUE TYPE
 - Bugfix Pull Request

##### SUMMARY
Addresses  https://github.com/ansible/ansible-container/issues/944
"Destroy command deletes all similar images of other projects too"


Voronenko:
There were no one to review, but useful in develop.

